PR #137, opened 2014-09-02
WIP: Photutils region extraction/photometry

keflavich:
This is almost a placeholder rather than a true WIP.  

Related: https://github.com/astropy/photutils/issues/121


keflavich:
This is effectively implemented in `subcube_from_regions` and similar.

PR #70, opened 2014-04-21
CLASS LMV reader

keflavich:
This is a reader for GILDAS-CLASS LMV cubes.  They are nearly identical to FITS cubes, just laid out in such a way that they have to be read differently.

I have no plans to write a writer for this format.

I have virtually no test data available for it either, but I'll try to generate some.

PR #159, opened 2014-11-14
LMV reader: support CLASS Type 2 files

keflavich:
Still, all the data I have access to are proprietary... I'm sure there's some public stuff somewhere.

This PR implements the "Version 2" CLASS LMV files.  The format is pretty easy to parse, and there are a lot of internal consistency checks too.


keflavich:
@vilhelmp - Responding to your [comment](https://github.com/radio-astro-tools/spectral-cube/pull/70#issuecomment-69310865) on another thread, could you send me a CLASS type 2 lmv cube too?  Ideally, something the same tiny size - 3x4x5 or something similar.  I have a basic test set up to work for CLASS type 1 cubes now (in this PR).  It doesn't work perfectly, unfortunately - there is an inconsistency in how the projections are translated that I'll have to debug.

for my own reference: the CLASS-exported version has `CTYPE1 = 'RA---ARC'`, but the projection type dictionary I've used results in `CTYPE1 = 'RA---AZP'`, which are not equivalent.

vilhelmp:
@keflavich how do I check what version the file is in? I have some data from 2012, so maybe.

keflavich:
I think the first few characters of the file indicate it... but I have to look that up.  v2 CLASS files are very new, mid-2014 and later.

vilhelmp:
Aha, well then I don't have any. Would it work if I re-save a v1 file?

keflavich:
I... don't know.   

Could you let me know how you cropped your cube in CLASS?  I have some v2 files that are just too big, and I don't know how to use CLASS to snip them down to a reasonable size

vilhelmp:
I used MAPPING, not CLASS, just viewed the maps in the window (go view) then export the cube.  Perhaps there is a similar command in CLASS when you view spectra, e.g. some help on "go browse".

keflavich:
huh, ok.  I've never heard of it.  Well, I'll ask some local experts.  Thanks for your effort!

PR #180, opened 2015-01-06
BUGFIX: moments on cubes with units.  Added regression test

keflavich:
I discovered a bug when trying to measure moments for cubes with units.  In short, we were using unit'd functions within _moment code when we should not have been.  I am not really sure what triggered the failures in my code, but I think it was pvextractor.  Anyway, the regression test now checks for it.


keflavich:
cc @ChrisBeaumont

PR #184, opened 2015-02-10
Parse JY/BEAM unit appropriately

keflavich:
This is a hack right now,  but it's a hack in the right direction: the Jy/Beam unit should actually have a RadioBeam as the unit.  

The tricky thing now is that something with units `u.Jy/beam`, where `beam` is a radio beam, should be convertible to Kelvin with only the frequency, NOT the beam area, being specified... and that's not possible right now.

cc @astrofrog, this is pertinent to other recent conversations


keflavich:
There's also the problem that the radio beam "unit" is not a unit but a quantity.  That breaks a lot of important things, like the `to_string` method.

spectral_cube/spectral_cube.py
```diff
@@ -237,11 +241,28 @@ def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
         # Deal with metadata first because it can affect data reading
         self._meta = meta or {}
         if 'BUNIT' in self._meta:
-            try:
-                self._unit = u.Unit(self._meta['BUNIT'])
-            except ValueError:
-                warnings.warn("Could not parse unit {0}".format(self._meta['BUNIT']))
-                self._unit = None
+
+            # special case: CASA makes non-FITS-compliant jy/beam headers
```
keflavich: it may now be best to split out the metadata parsing into a separate function or even separate .py file

jpinedaf:
The current implementation works for ``BUNIT == JY/BEAM`` but not for ``BUNIT == Jy/beam``, right?

keflavich:
@jpinedaf actually the current implementation doesn't really work and reveals an underlying error in the tests.  I'm working on it.

keflavich:
@jpinedaf now this should work.  It's best if you have [`radio_beam`](https://github.com/radio-astro-tools/radio_beam) installed.

Note that this is a hack and should be fixed once we have some solution for "beam" as a unit in astropy.  @astrofrog, any comments?  Could you also review this?

keflavich:
I can't reproduce the failures locally.

keflavich:
I can't reproduce these errors *remotely* either.  Either there is an error with the `.travis.yml` file or with travis itself.

astrofrog:
There are several issues:

* A couple of test failures at this line:

```
>               out[x, y] = function(data, **kwargs)
```

Maybe add debug print statements just before to see why that might be

* Radio-beam is using ``execfile`` (it really shouldn't!) which I think is called something different in Python 3

keflavich:
@astrofrog - regarding the first, yes, I can add some print functions, that's reasonable enough. But, can you reproduce the failure locally?

astrofrog:
I'll check now

astrofrog:
Yes I can reproduce the issue, will investigate.

astrofrog:
This is the failure that's happening:

```
In [1]: import numpy as np

In [2]: from astropy import units as u

In [3]: x = np.ones((2,2))

In [4]: x[0,0] = 4 * u.K
ValueError                                Traceback (most recent call last)
<ipython-input-4-1edabbf385e2> in <module>()
----> 1 x[0,0] = 4 * u.K

ValueError: setting an array element with a sequence.
```

Seems to be an astropy bug (the unit should just be dropped I think). The workaround is to write:

```
                result = function(data, **kwargs)
                if isinstance(result, u.Quantity):
                    out[x, y] = result.value
                else:
                    out[x, y] = result
```

(on line 588 of spectral_cube). I'll report on astropy.

keflavich:
Thanks!  There ought to be a fix in spectral-cube too, I'm just perplexed as to why I don't get it locally.

astrofrog:
What numpy/astropy versions do you have?

astrofrog:
I have Numpy 1.9.2 and Astropy 1.0.3

astrofrog:
I also see the radio_beam install failure with Python 3:

```
$ pip install https://github.com/rdio-astro-tools/radio_beam/archive/master.zip
You are using pip version 7.0.3, however version 7.1.0 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting https://github.com/radio-astro-tools/radio_beam/archive/master.zip
  Downloading https://github.com/radio-astro-tools/radio_beam/archive/master.zip
     - 937kB 1.2MB/s
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/pip-tw_1_jcq-build/setup.py", line 28, in <module>
        execfile('radio_beam/version.py')
    NameError: name 'execfile' is not defined
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /var/folders/zy/t1l3sx310d3d6p0kyxqzlrnr0000gr/T/pip-tw_1_jcq-build
```

I think that package shouldn't be using ``execfile``. Maybe it should be updated to use the astropy template?

keflavich:
Yes, radio-beam needs to be updated to use the package template.  I don't have time for that right now.

My versions:
```
$ python -c "import numpy; print numpy.__version__; import astropy; print astropy.__version__"
1.9.2
1.1.dev12825
```
so maybe it's an astropy-dev issue?  

My goal is to create a regression test for this too.  I'll try testing against astropy-1.0.3 to see if I can reproduce now..

astrofrog:
@keflavich - at least radio_beam will need to be updated to not use execfile but the Python 3 equivalent when using Python 3: http://stackoverflow.com/questions/436198/what-is-an-alternative-to-execfile-in-python-3-0

keflavich:
Aha, finally, a green!  @astrofrog unless you have objections I'll go ahead with this.

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -34,6 +34,12 @@
 except ImportError:
     bottleneckOK = False
 
+try:
+    import radio_beam
+    radiobeamOK = True
```
astrofrog: We normally use ``RADIOBEAM_INSTALLED`` convention (I like installed better than OK)

astrofrog:
Apart from my minor comment above, good to go!

keflavich:
....coveralls is a little late to the party?

PR #186, opened 2015-02-12
Fix air wavelength conversion

keflavich:
Air wavelengths were being treated incorrectly due to mishandling of the CTYPE keywords.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@
 - Output mask as a CASA image https://github.com/radio-astro-tools/spectral-cube/pull/171
 - ytcube exports to .obj and .ply too
   https://github.com/radio-astro-tools/spectral-cube/pull/173
+- Fix air wavelengths, which were mistreated (https://github.com/radio-astro-tools/spectral-cube/pull/186)
 
 0.2.1 (2014-12-03)
 ------------------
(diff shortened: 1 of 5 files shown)


keflavich:
@astrofrog This PR fixed the issue I mentioned yesterday.  It was a really, really stupid bug that I missed because there was no test coverage for it.  However, with the new tests, I've been forced to use some relatively large tolerances - I am not sure the accuracy of the air wavelength conversions is better than 10^-6, which I'm concerned may not be enough for some use cases (e.g., R~100,000, that's 10% of a pixel).  How does astropy deal with indicating accuracy limitations, e.g. in time?

astrofrog:
@keflavich - I'm not sure how/whether ``Time`` deals with this kind of issue, but maybe @mhvk or @taldcroft can comment? (in particular, I think it's possible that ``Time`` doesn't *always* do full precision operations, but I might be wrong)

mhvk:
`Time` has no specific mechanism to indicate uncertainty. I do think this is something we should eventually have in `astropy`, but I think it needs a general uncertainty framework.

As a comment on this issue: @keflavich - if you need to set what you think are unduly large tolerances, I think it would be good to raise a separate issue about it. Often, with a bit of digging, one can improve, but if we have no record that it might be an issue, we'll never do it!

PR #188, opened 2015-02-15
Return OneDSpectrum when appropriate

keflavich:
The `OneDSpectrum` class was implemented previously to provide usable slices of the form `cube[:,y,x]`, but the related numpy functions were never updated to provide similar access.  Now they do.

We are close to the point of providing users with dangerous rope, though: cubes can easily be in surface brightness units, which make sense to average but lose meaning when summed unless the pixel size (pixels-per-beam) is accounted for.  We're still no worse than any other software, but an important goal is to seamlessly incorporate the `radio_beam` information so that "summed" or "averaged" spectra really have the right units.


keflavich:
This PR has revealed that `spectral_cube` requires `numpy>=1.8` to use the `.nan(mean|std)` functions and `axis=(1,2)` syntax.  I'm therefore deprecating support for older versions.  Most of `spectral_cube` will still work with these, but I don't think it is essential to support these in a beta-ish package like this.

keflavich:
@astrofrog @ChrisBeaumont any comments on the loss of np < 1.8?

keflavich:
We can add tests back in later, but they will require some `pytest.skipif` marks

ChrisBeaumont:
Fine by me

On Sunday, February 15, 2015, Adam Ginsburg <notifications@github.com>
wrote:

> Merged #188 <https://github.com/radio-astro-tools/spectral-cube/pull/188>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/radio-astro-tools/spectral-cube/pull/188#event-236019759>
> .
>


-- 
*************************************
Chris Beaumont
Senior Software Engineer
Harvard Center for Astrophysics
60 Garden Street, MS 42
Cambridge, MA 02138
chrisbeaumont.org
*************************************

astrofrog:
I'm also fine with requiring Numpy 1.8+

PR #191, opened 2015-03-09
add __eq__ and __ne__ comparisons (which yield lazy masks)

keflavich:
These were just overlooked, I think.  Previously, `__eq__` and `__ne__` would use `object`'s equality, which I think is identical to `is`, e.g.:

```
x = object()
x == x
x is x
```
those comparisons should be the same, afaik

cc @astrofrog - this is a nearly trivial change; can you think of any reason not to incorporate it?

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -1596,6 +1596,9 @@ class SpectralCube(object):
     def __eq__(self, value):
         return LazyMask(lambda data: data == value, data=self._data, wcs=self._wcs)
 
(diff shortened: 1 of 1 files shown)


keflavich:
Apparently python3 doesn't like `__eq__`?

ChrisBeaumont:
I think this is related to the advice that, on python3, you should always implement `__eq__ and __hash__` together (I don't know why)

keflavich:
thanks @ChrisBeaumont , adding `__hash__` solved it.

PR #194, opened 2015-03-13
Beam information lost in moment maps

keflavich:
Hi,

I am calculating the integrated intensity map in a cube:
```file_in='NGC1333_ABCDEFGH_NH3_11_all.fits'
cube_raw = SpectralCube.read(file_in)
cube = cube_raw.with_spectral_unit(u.km / u.s,velocity_convention='radio')
mom=cube.moment(order=0)
mom.write( 'NGC1333_test_Tdv.fits')
```
that has the following beam information in the raw cube
```
print repr(cube.header)
SIMPLE  =                    T / Written by IDL:  Sat Feb  7 00:02:58 2015      
BITPIX  =                  -32 / Number of bits per data pixel                  
NAXIS   =                    3                                                  
NAXIS1  =                  192                                                  
NAXIS2  =                  336                                                  
NAXIS3  =                  801                                                                   
----- snip
BMAJ    =           0.00870000 /                                                
BMIN    =           0.00870000 /                                                
BPA     =        0.00000000000 /                                                
----- snip                                               
```
while the full header if the integrated intensity map does not have the beam information (it lost a few keywords):
```
In [27]: print repr(hdulist[0].header)
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                  -64 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                  192                                                  
NAXIS2  =                  336                                                  
WCSAXES =                    2 / Number of coordinate axes                      
CRPIX1  =                107.0 / Pixel coordinate of reference point            
CRPIX2  =                168.0 / Pixel coordinate of reference point            
CDELT1  =              -0.0029 / [deg] Coordinate increment at reference point  
CDELT2  =               0.0029 / [deg] Coordinate increment at reference point  
CUNIT1  = 'deg'                / Units of coordinate increment and value        
CUNIT2  = 'deg'                / Units of coordinate increment and value        
CTYPE1  = 'RA---TAN'           / Right ascension, gnomonic projection           
CTYPE2  = 'DEC--TAN'           / Declination, gnomonic projection               
CRVAL1  =        52.2893333435 / [deg] Coordinate value at reference point      
CRVAL2  =        31.3283596039 / [deg] Coordinate value at reference point      
LONPOLE =                180.0 / [deg] Native longitude of celestial pole       
LATPOLE =        31.3283596039 / [deg] Native latitude of celestial pole        
RESTFRQ =        23694495500.0 / [Hz] Line rest frequency                       
WCSNAME = 'LSRK-Freq'          / Coordinate system title                        
RADESYS = 'FK5'                / Equatorial coordinate system                   
EQUINOX =               2000.0 / [yr] Equinox of equatorial coordinates         
SPECSYS = 'LSRK'               / Reference frame of spectral coordinates        
BUNIT   = 'K km s-1'         
```

Do I really have to restore the beam information manually after the moment map is calculated?

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -98,13 +98,22 @@ class LowerDimensionalObject(u.Quantity):
         return self._meta
 
     @property
+    def header(self):
+        header = self._header
+        # This inplace update is OK; it's not bad to overwrite WCS in this
+        # header
+        if self.wcs is not None:
+            header.update(self.wcs.to_header())
+        header['BUNIT'] = self.unit.to_string(format='fits')
+        header.insert(2, Card(keyword='NAXIS', value=self.ndim))
+        for ind,sh in enumerate(self.shape[::-1]):
+            header.insert(3+ind, Card(keyword='NAXIS{0:1d}'.format(ind+1),
+                                      value=sh))
+        return header
+
+    @property
     def hdu(self):
-        from astropy.io import fits
-        if self.wcs is None:
-            hdu = fits.PrimaryHDU(self.value)
-        else:
-            hdu = fits.PrimaryHDU(self.value, header=self.wcs.to_header())
-        hdu.header['BUNIT'] = self.unit.to_string(format='fits')
+        hdu = PrimaryHDU(self.value, header=self.header)
         return hdu
 
     def write(self, filename, format=None, overwrite=False):
@@ -131,7 +140,7 @@ class LowerDimensionalObject(u.Quantity):
 class Projection(LowerDimensionalObject):
 
     def __new__(cls, value, unit=None, dtype=None, copy=True, wcs=None,
-                meta=None, mask=None):
+                meta=None, mask=None, header=None):
 
         if np.asarray(value).ndim != 2:
             raise ValueError("value should be a 2-d array")
@@ -144,6 +153,10 @@ class Projection(LowerDimensionalObject):
         self._wcs = wcs
         self._meta = meta
         self._mask = mask
+        if header is None:
+            self._header = header
+        else:
+            self._header = Header()
 
         return self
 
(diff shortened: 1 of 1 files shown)


keflavich:
@jpinedaf I want to add some tests before I merge this, but it should solve your problem.  All header keywords except WCS keywords will now be preserved from the original file

astrofrog:
I haven't had a chance to check over this, but this should probably only be preserved if the moment map is along the spectral axis, since the beam won't mean anything in PV space?

keflavich:
Actually, that's not true: the beam is just as meaningful in PV space, since the beam governs the conversion between Jy and K.

low-sky:
An integration in the spatial direction makes surface brightness units into ... linear brightness.  I see the point that retaining the information is useful for making the conversion, but once you've integrated a surface brightness unit by a dimension that carries sqrt(sr)=radian units, the resulting value shouldn't be expressed in K or Jy/beam.  Jy would be cleanly integrable. 

Carrying the beam really only provides context for the expert user in the dimensional post-mortem but assuming that header info then applies to the data makes little sense.  You end up in the optical spectroscopy domain where the spatial PSF is only characterized in one dimension.  Is there a keyword for that?

keflavich:
True, integrating in the spatial domain does give up information.   But
slicing or interpolating would not.   So slices could keep the info but
moments/sums/means should drop it?

keflavich:
In the currently implemented solution, ALL header information except strictly WCS information is preserved.  As I see it right now, we have two options: 

(1) Preserve none of the header information
(2) preserve all of the header information

The hybrid I had somewhat proposed - maintain beam information iff projected along the spectral axis - is not possible to implement prior to complete integration of `radio_beam`.

So, is there any harm in preserving the non-WCS header keywords?  I would normally say "no, keep everything".  If anything, we should be adding provenance information, e.g. the original FITS filename and commands used to get to the current slice/projection, if possible

keflavich:
I think having beam information plus potentially uninformative information is better than not having beam information.  I'm therefore merging.

PR #195, opened 2015-03-16
Added method to make channel maps

astrofrog:
Not sure if this would be useful to have here, but example usage:

```
import numpy as np
cube = SpectralCube.read('HD163296_13CO.image.fits')
cube.plot_channel_maps('test.png', 6, 6, channels=range(10, 46),
                       levels=np.linspace(0.05, 0.5, 10), zoom=2)
```

test

Does this belong here or in e.g. WCSAxes?

If here, still needs:
- [ ] tests
- [ ] docs
- [ ] mention that WCSAxes is needed
- [ ] a way to set unit for spectral axis labels

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -1846,6 +1846,67 @@ class SpectralCube(object):
         hdu = fits.PrimaryHDU(self.filled_data[:].value, header=self.header)
         return hdu
 
+    def plot_channel_maps(self, output_file, nx, ny, channels, levels,
+                          figsize=(20,20), decimals=3, zoom=1):
+        """
+        Make channel maps from a spectral cube
+
+        Parameters
+        ----------
+        input_file : str
+            Name of the input spectral cube
+        output_file : str
+            Name of the matplotlib plot
+        nx, ny : int
+            Number of sub-plots in the x and y direction
+        channels : list
+            List of channels to show
+        levels : list
+            List of levels to show
+        figsize : tuple, optional
+            Figure size for matplotlib
+        decimals : int, optional
+            Number of decimal places to show in spectral value
+        zoom : int, optional
+            How much to zoom in. In future versions of this function, the
+            pointing center will be customizable.
+        """
+
+        import matplotlib.pyplot as plt
+
+        if len(channels) != nx * ny:
+            raise ValueError("Number of channels should be equal to nx * ny")
+
+        # Read in spectral cube and get spectral axis
+        spectral_axis = self.spectral_axis
+
+        fig = plt.figure(figsize=figsize)
+
+        sizey, sizex = self.shape[1:]
+        cenx = sizex / 2.
+        ceny = sizey / 2.
+
+        for ichannel, channel in enumerate(channels):
+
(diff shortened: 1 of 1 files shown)


keflavich:
This is pretty easy to do without a dedicated method, except perhaps the grid setup method.  I think this may be more appropriate as an annotated example.  However, I don't really object to the method being implemented with the various caveats you suggest.

astrofrog:
@keflavich - maybe this actually belongs in APLpy, if it is to become a convenience layer on top of WCSAxes. It would certainly be a more natural place for people to find it.

keflavich:
@astrofrog - it definitely belongs in APLpy.  It is a nice quicklook-style feature for `spectral_cube`, but it is limited: it would pretty much have to raise an exception for cubes with >100 channels, and probably ought to warn loudly above 25 channels, since those could take forever to plot.

This also reminds me that we need to implement & include downsampling routines in spectral_cube.  I have some in FITS_tools, I think, but they definitely belong here and can probably be implemented with a cleaner strategy and a more appropriate treatment of WCS.

keflavich:
@astrofrog despite my comments earlier, I think we still want this.  It would also be cool to do the PV version of this.

spectral_cube/spectral_cube.py
```diff
@@ -1846,6 +1846,67 @@ def hdu(self):
...
+            Number of decimal places to show in spectral value
+        zoom : int, optional
+            How much to zoom in. In future versions of this function, the
+            pointing center will be customizable.
+        """
+
+        import matplotlib.pyplot as plt
+
+        if len(channels) != nx * ny:
+            raise ValueError("Number of channels should be equal to nx * ny")
```
keflavich: Perhaps also add `if len(channels) > 100 and not allow_huge: raise ValueError("Large number of subplots!  Specify allow_huge=True if you really want this."`

spectral_cube/spectral_cube.py
```diff
@@ -1846,6 +1846,67 @@ def hdu(self):
...
+
+        # Read in spectral cube and get spectral axis
+        spectral_axis = self.spectral_axis
+
+        fig = plt.figure(figsize=figsize)
```
keflavich: should allow the figure object to be passed in, or default to using `plt.gcf()`.  But it is important to specify figsize... not sure about the best API here.

spectral_cube/spectral_cube.py
```diff
@@ -1846,6 +1846,67 @@ def hdu(self):
...
+            ax.set_ylim(ceny - ceny / zoom, ceny + ceny / zoom)
+            ax.set_title(("{0:." + str(decimals) + "f}").format(spectral_axis[channel]))
+
+            if ichannel % nx != 0:
+                ax.coords[1].set_ticklabel_position('')
+
+            if ichannel < nx * (ny - 1) != 0:
+                ax.coords[0].set_ticklabel_position('')
+
+        fig.savefig(output_file)
```
keflavich: can `output_file` be optional?

keflavich:
@astrofrog could you make this editable by maintainers?  I want to start pushing some changes to it

keflavich:
Will be closed by #551

PR #196, opened 2015-04-09
Preserve BUNIT in header when it cannot be parsed

keflavich:
e.g., JY/BEAM should stay in the output header even though there is no appropriate Unit for it


keflavich:
If `coveralls` tells me I reduced coverage again I swear I will hunt it down and cut its suspenders...

keflavich:
oh it is ON.

PR #198, opened 2015-04-26
Improved __repr__

keflavich:
Improve the cube `__repr__` by showing the extrema in values.  This involves caching the min/max of each dimension the first time it is called.  Since we have treated cubes as immutable, this should be the right approach.

@ChrisBeaumont - not sure you'll have time to check this, but could you lay out somewhere what you had in mind for `world_spines`?  That would be a more efficient way to check the extrema

@astrofrog, @ChrisBeaumont - any concerns about performance here?  I think it's safe to assume that each axis is small enough for this process in general, though it might *not* be safe to assume that a full plane is cheap, hence the need for `world_spines`.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -366,22 +366,22 @@ class SpectralCube(object):
     @property
     @cached
     def spectral_extrema(self):
-        self._spectral_min = self.spectral_axis.min()
-        self._spectral_max = self.spectral_axis.max()
+        _spectral_min = self.spectral_axis.min()
+        _spectral_max = self.spectral_axis.max()
 
-        return self._spectral_min, self._spectral_max
+        return _spectral_min, _spectral_max
 
     @property
     @cached
     def world_extrema(self):
         lat,lon = self.spatial_coordinate_map
-        self._lon_min = lon.min()
-        self._lon_max = lon.max()
-        self._lat_min = lat.min()
-        self._lat_max = lat.max()
+        _lon_min = lon.min()
+        _lon_max = lon.max()
+        _lat_min = lat.min()
+        _lat_max = lat.max()
 
-        return ((self._lon_min, self._lon_max),
-                (self._lat_min, self._lat_max))
+        return ((_lon_min, _lon_max),
+                (_lat_min, _lat_max))
 
     @property
     @cached
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -363,6 +363,7 @@ class SpectralCube(object):
                                              ))
         return s
 
+    @property
     @cached
     def spectral_extrema(self):
         self._spectral_min = self.spectral_axis.min()
@@ -370,6 +371,7 @@ class SpectralCube(object):
 
         return self._spectral_min, self._spectral_max
 
+    @property
     @cached
     def world_extrema(self):
         lat,lon = self.spatial_coordinate_map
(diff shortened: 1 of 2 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -340,11 +340,64 @@ def __repr__(self):
...
+        s += (" n_s: {0:6d}  type_s: {1:8s}  unit_s: {2:5s}"
+              "  range: {3:8s}:{4:8s}".format(self.shape[0],
+                                              self.wcs.wcs.ctype[2],
+                                              self.wcs.wcs.cunit[2],
+                                              self.spectral_extrema[0],
+                                              self.spectral_extrema[1],
+                                             ))
         return s
 
+    @property
```
ChrisBeaumont: How do you feel about making a `cached_property` decorator, instead of checking for all these `hasattr` checks?

spectral_cube/spectral_cube.py
```diff
@@ -340,11 +340,64 @@ def __repr__(self):
...
+    def spectral_extrema(self):
+        if not hasattr(self, '_spectral_min'):
+            self._spectral_min = self.spectral_axis.min()
+        if not hasattr(self, '_spectral_max'):
+            self._spectral_max = self.spectral_axis.max()
+
+        return self._spectral_min, self._spectral_max
+
+    @property
```
ChrisBeaumont: same comment here

ChrisBeaumont:
IIRC the intention of world spines was to compute 1D arrays for each of the world coordinates, which could be broadcast to the full world coordinates arrays. This would be more memory efficient in situations where the pix<->world transformation is separable along each axis (like cartesian coordinates). I don't think we implemented it because of the wrinkle of identifying when a particular WCS can be represented via N broadcast 1D arrays, instead of N ND arrays.

spectral_cube/spectral_cube.py
```diff
@@ -340,11 +340,59 @@ def __repr__(self):
...
+                                              self.wcs.wcs.cunit[2],
+                                              self.spectral_extrema[0],
+                                              self.spectral_extrema[1],
+                                             ))
         return s
 
+    @property
+    @cached
+    def spectral_extrema(self):
+        self._spectral_min = self.spectral_axis.min()
```
ChrisBeaumont: There's probably no reason to store these results anymore, right?

spectral_cube/spectral_cube.py
```diff
@@ -340,11 +340,59 @@ def __repr__(self):
...
+        return self._spectral_min, self._spectral_max
+
+    @property
+    @cached
+    def world_extrema(self):
+        lat,lon = self.spatial_coordinate_map
+        self._lon_min = lon.min()
+        self._lon_max = lon.max()
+        self._lat_min = lat.min()
+        self._lat_max = lat.max()
```
ChrisBeaumont: same here

spectral_cube/spectral_cube.py
keflavich: in this case, yes

ChrisBeaumont:
This looks good to me.

One followup issue (perhaps for another PR): the current implementation of `cached` will prevent any instance of SpectralCube from being garbage collected (because it's holding `self` instances in the cache dictionary). We probably want to avoid doing that.

keflavich:
what the heck does this mean?
```
  File "<stdin>", line 38
    print(darkred('reST markup error:'), file=sys.stderr)
                                             ^
SyntaxError: invalid syntax
Sphinx Documentation subprocess failed with return code 1
```

ChrisBeaumont:
python3 print syntax run on python2 by any chance?

keflavich:
umm... interesting thought.  I have no idea where, though.  Certainly not in this PR.

PR #199, opened 2015-05-04
Improvement of quick look

keflavich:
It would be great if the quick look for spectra is shown in "histogram" mode instead of the current, "connect the dots" mode.


keflavich:
@jpinedaf a solution is in place now

PR #202, opened 2015-05-05
Consistently decode strings from lmv files

sargas:
A line like `header['CTYPE{0}'.format(ii+1)] = _ctype_dict[ctype.strip()]` doesn't work too well when `_ctype_dict` has utf8 keys and `ctype` is not decoded. I'm not very familiar with lmv files to generate a test case (and the file that hits this problem isn't mine to share).

This change just makes sure that any strings read are decoded and stripped before used.

### DIFF
--- a/spectral_cube/io/class_lmv.py
+++ b/spectral_cube/io/class_lmv.py
@@ -59,7 +59,7 @@ def read_lmv(filename):
 
     with open(filename,'rb') as lf:
         # lf for "LMV File"
-        filetype = lf.read(12).decode('utf8')
+        filetype = _read_string(lf, 12)
         #!---------------------------------------------------------------------
         #! @ private
         #!       SYCODE system code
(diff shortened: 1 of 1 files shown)


keflavich:
@sargas This mostly looks good, but is it really necessary to remove the `.strip` on, e.g., line 484?  It seems like that could result in whitespace comparison failures.

If you can find any (small) lmv files to add to the tests, they would be very welcome.

sargas:
In line 484, `code` is from the iteration over `ijcode`, which is defined in the line
`ijcode = [_read_string(lf, 12) for ii in range(gdf_maxdims)]`. The `_read_string` function strips the values, so another `strip` has no effect.

Is there a way to zero out a spectral cube and change its headers to anonymous-ize it?

keflavich:
@sargas - for an LMV cube, unfortunately, I don't know how to do that reliably.  I've implemented an LMV reader, not a writer...

keflavich:
The single build failure is from #203 and unrelated.

PR #209, opened 2015-06-03
Split out lower-dimensional objects

keflavich:
split out projection, slice, and onedspectrum into their own module. We'll
probably want to import these from somewhere else in the long term anyway


keflavich:
The "red" is just coveralls being a doofus.  There was no change in real coverage.

astrofrog:
btw you can now configure coveralls to have a threshold for a build to fail (e.g. -0.01%)

keflavich:
well, in this case, it's just wrong (afaict) - I didn't add or remove any lines of code.... though maybe the import statements count?

PR #210, opened 2015-06-07
Add a splatalogue-based line-search utility

keflavich:
Search the spectral_axis range for lines using splatalogue, with the option of
a velocity shift.  

cc @teuben - very basic, but this is a first step for ADMIT, no?


keflavich:
This requires a remote-data test...

PR #211, opened 2015-06-13
Incorrect results from cube.std(axis=0) with 'slice' iteration strategy

keflavich:
See https://github.com/jpinedaf/GAS/issues/34#issuecomment-111629067 for details.

This would only affect large cubes because of the automatic iterator strategy selected.  It also raises a 'todo' item, which is to implement a one-pass slice-by-slice standard deviation.  That's not difficult but requires testing, and it might not be much faster.

Also, we never implemented ray-iteration strategies for numpy functions, which might be faster for sparse cubes like this.  We should cache the 1-dimensional mask include...

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -431,13 +431,17 @@ class SpectralCube(object):
 
         projection = self._naxes_dropped(axis) in (1,2)
 
+        if how == 'slice':
+            raise NotImplementedError("Standard deviation cannot be computed "
+                                      "in a slicewise manner.  Please use a "
+                                      "different strategy.")
+
         # standard deviation cannot be computed as a trivial step-by-step
         # process.  There IS a one-pass algorithm for std dev, but it is not
-        # implemented, so we must force cube here by specifying reduce=False.  We could and should also
+        # implemented, so we must force cube here.  We could and should also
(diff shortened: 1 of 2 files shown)


keflavich:
@jpinedaf this is the solution to your bug

@ChrisBeaumont we found some serious bugs in the slice-iteration reduction approach that hid for a long time because we rarely tried this on huge cubes and the tests didn't cover the huge-cube case.  I've gotten around this by iterating over the used strategies.  

`mean` and `stddev` will be slow and memory-intensive for huge cubes, and we can get past this by implementing an appropriate one-pass solution, but that's a task for another day.

ChrisBeaumont:
@keflavich interesting, did you isolate the cause of the bug itself?

keflavich:
yes - the standard deviation and mean were computed in a running fashion just like sum and min/max, etc.  e.g.:

```
x = [1,2,3,4]
running_mean(x) = [mean(1,2), mean((mean(1,2), 3), mean(mean((mean(1,2), 3), 4)]
```

which is wrong.  In order to compute a running mean/standard deviation, a different algorithm needs to be implemented.  This only affected the slice iteration, which only came into effect for npoints > 1e8

ChrisBeaumont:
ah, right. good catch.

PR #217, opened 2015-07-03
For comparison operations, check units

keflavich:
When comparing to a value, if it's a quantity, convert to the cube's unit
first.  Needed to be done for all comparison operations.

NOTE that there is an outstanding bug: the "value" placed into the lamda function
does not get sliced when the rest of the SpectralCube gets sliced.


keflavich:
This is now built on top of `jybeam` since there are some redundant changes.

I don't know what to do about the travis failures, since I can't reproduce them.

Issue #193, opened 2015-03-13
Problem with example for yt write_bitmap


esoPanda:
In this example:

http://spectral-cube.readthedocs.org/en/latest/yt_example.html#visualization-example

the final step, `write_bitmap(snapshot, 'cube_rendering.png', transpose=True)`, fails with this traceback:
```
In [72]: write_bitmap(snapshot, 'cube_rendering.png', transpose=True)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-72-d570ef0f1c1c> in <module>()
----> 1 write_bitmap(snapshot.value, 'cube_rendering.png', transpose=True)

/home/kwang/anaconda/lib/python2.7/site-packages/yt/visualization/image_writer.pyc in write_bitmap(bitmap_array, filename, max_val, transpose)
    165         bitmap_array = bitmap_array.swapaxes(0,1)
    166     if filename is not None:
--> 167         pw.write_png(bitmap_array, filename)
    168     else:
    169         return pw.write_png_to_string(bitmap_array.copy())

/home/kwang/anaconda/lib/python2.7/site-packages/yt/utilities/png_writer.pyc in write_png(buffer, filename, dpi)
     17     width = buffer.shape[1]
     18     height = buffer.shape[0]
---> 19     _png.write_png(buffer, width, height, filename, dpi)
     20 
     21 def write_png_to_string(buffer, dpi=100, gray=0):

TypeError: First argument must be an rgba buffer.
> /home/kwang/anaconda/lib/python2.7/site-packages/yt/utilities/png_writer.py(19)write_png()
     18     height = buffer.shape[0]
---> 19     _png.write_png(buffer, width, height, filename, dpi)
     20 

ipdb> 

```

@jzuhone can you comment on this?

jzuhone:
Hi @esoPanda, just a quick question first--what's the version of yt you are using?

esoPanda:
I'm using 3.0.2

esoPanda:
@jzuhone, I'm using yt 3.0.2

jzuhone:
Ok, I'll have a look at this later today and see if I can reproduce it.

keflavich:
@jzuhone when we inspected the data, `buffer` was a `[N,N,4]` Quantity (so it had a unit).  I tried using the `.value` to get a pure numpy dtype=float64 array and that did not solve the problem.

jzuhone:
[...]

PR #219, opened 2015-07-16
Fixing yt volume rendering script in docs

jzuhone:
This updates the script in the docs to make a volume rendering with yt to be compatible with current API.

NOTE: This does _not_ fix [Issue 193](https://github.com/radio-astro-tools/spectral-cube/issues/193), that is a bug on the yt side that is fixed [here](https://bitbucket.org/jzuhone/yt/commits/f2357819d9dcbdcdb3cdd830c04dd7bb45a31290), which will be merged into yt today (hopefully) and be included in the next stable release (3.2).

### DIFF
--- a/docs/yt_example.rst
+++ b/docs/yt_example.rst
@@ -68,12 +68,12 @@ produce a 3-d isocontour visualization using an object returned by
 :meth:`~spectral_cube.SpectralCube.to_yt`::
 
     import numpy as np
-    from spectral_cube import read
+    from spectral_cube import SpectralCube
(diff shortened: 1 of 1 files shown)


keflavich:
I don't know what's causing the python3 failures and I can't reproduce them locally, so I might ignore them...

jzuhone:
Do the example scripts in the docs get run as tests?

keflavich:
...good question.  Even if they do, though, your changes shouldn't have affected py3.

And after restarting the tests, they don't.  I don't know what went wrong, but it was ephemeral.

PR #220, opened 2015-07-17
Apply everywhere: applying functions to whole cube

keflavich:
This PR resumes #101 and follows #100.  It is intended as a first step to addressing a problem noted on the python in astronomy facebook group:
```
I'm trying to apply simple arithmetic operations to spectral cubes using SpectralCube, how can I do this to the data cube itself without extracting the data into numpy arrays? I'd like to maintain the data cube format.
```

Todo:
 * [x] Add tests
 * [x] Implement basic arithmetic functions (`add`, `subtract`, `product`, `divide`, `pow`)
 * [x] Add documentation

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,10 @@
    (https://github.com/radio-astro-tools/spectral-cube/pull/210)
  - Bugfixes (211,212,217)
  - Add arithmetic operations (add, subtract, divide, multiply, power)
-   (https://github.com/radio-astro-tools/spectral-cube/pull/220)
+   (https://github.com/radio-astro-tools/spectral-cube/pull/220).
+   These operations will not be permitted on large cubes by default, but will
+   require the user to specify that they are allowed using the attribute
+   ``allow_huge_operations``
 
 0.2.2 (2015-03-12)
 ------------------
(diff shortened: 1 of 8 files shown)


keflavich:
@ChrisBeaumont @astrofrog this represents a major change to the SpectralCube UI.  I'd like a review, or at least a :+1: or `wontread` from each of you before merging.  Merging this should trigger a new release - 0.3?

keflavich:
I'll merge this by the end of the week if I don't hear from anyone.

ChrisBeaumont:
Changes look ok to me, though this will make it easier to trigger memory
errors.

keflavich:
Good point.  Maybe we should raise an Exception instead of a warning, but allow users to override that with a ConfigurationItem?

ChrisBeaumont:
It's a judgement call - maybe arithmetic is sufficiently fundamental that
it's worth it

Since the workshop, blaze and xray have matured quite a bit with respect to
the kind of lazy, out of core math in this library. At some point it might
be worth investigating whether it's worth it to implement spectral cube on
top of those data structures instead of ndarrays.

keflavich:
@ChrisBeaumont I think you're wise to advocate not allowing all-in-memory operations by default, at least for large cubes, so I've added that: an exception will be raised if you try to perform arithmetic on large cubes unless the `allow_huge_operations` attribute is explicitly set.

keflavich:
... BUT I have had a long series of requests for this feature, and I've wanted it myself many times, so we should allow it.  Laziness would be great, but we'll have to come back to that later.

keflavich:
The failures are related to https://github.com/astropy/astropy/pull/3992, which means that to get tests passing we may have to remove some tests permanently.

PR #224, opened 2015-07-29
Allow for velocity coordinate system changes

keflavich:
A major feature request, and fundamentally this ought to be in specutils, but I want it here first.

The FITS standard defines many useful reference frames, and we should allow transformation between them.  TOPO/GEO->other frames will be hard, so we can put that off, but transformations between other frames are well-defined and easy, so those should be a higher priority.

This PR so far contains only a documentation skeleton for a new function `with_reference_frame`.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -826,8 +826,42 @@ class SpectralCube(object):
         """
         return self._new_cube_with(fill_value=fill_value)
 
+    def with_reference_frame(self, reference_frame, source_redshift=None,
+                             source_reference_frame=None):
+        """
+        Parameters
+        ----------
+        reference_frame : str
+            The frame of rest of the specified frequency, wavelength, or
+            velocity, FITS keyword SPECSYS.  Must be one of the allowed
+            keywords in FITS Paper III, Table 12: TOPOCENT, GEOCENTR, BARYCENT,
+            HELIOCEN, LSRK, LSRD, GALACTOC, LOCALGRP, CMBDIPOL, or SOURCE.
+            If SOURCE is used, the ``source_reference_frame`` and
+            ``source_redshift`` must also be specfied
+        source_reference_frame : str
+            The frame of rest of the specified frequency, wavelength, or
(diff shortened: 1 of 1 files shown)


astrofrog:
Actually I wonder if this could live in ``astropy.wcs.utils`` once it works?

keflavich:
Yes, that's probably a good place for it.  I keep being surprised that no one has implemented these things... I'm sure they exist in AST and in casacore, but... well.

astrofrog:
Tagging @eteq since this has some connections to what we discussed at pyastro15 - but the difference here is that this is to modify WCSes, rather than transform coordinates directly.

keflavich:
FYI, this came about because of a specific feature request from a colleague to be able to convert spectra (or cubes) into the `SOURCE` rest frame, so that's my highest priority.

eteq:
@keflavich @astrofrog - so just to be clear, this is about being able to take things in one WCS and convert to a *different* velocity frame, or just figure out how to interpret different input keywords?

I ask because the conversion I think *should* be implemented as part of a related coordinates effort.  But definitely that's not an easy thing to do right, so a temporary solution here limited only to the WCS-valid keywords is a reasonable interim approach.

keflavich:
@eteq - yes, this is about conversion.  Interpretation is kind of trivial, because if a file implements the FITS convention correctly, the conversion from pixel coordinates to world coordinates for any given WCS is unaffected by these various keywords, only transformations between different WCSs are affected.

eteq:
@keflavich - I wonder if wcslib will do these velocity transformations?  If so that might be an easy way to do it: let wcslib via `astropy.wcs` take care of it?

keflavich:
@eteq - as far as I can tell, it does not handle transformations; it instead treats these keywords as "auxiliary": e.g. http://www.atnf.csiro.au/people/mcalabre/WCS/wcslib/structwcsprm.html#ac089e5d0e3191255ceaea7f8591b27ea.  However, @mdboom, could you comment further - can wcslib compute transformations between different frames and/or different `CTYPE`s?

keflavich:
Closing: this will be handled by specutils in https://github.com/astropy/specutils/issues/422

PR #225, opened 2015-08-01
Slicewise standard deviation & mean

keflavich:
*WIP: expected to fail*
This is some initial work on a streaming standard deviation calculation to
allow for slicewise reduction (also for mean).  Right now, it only "works" in
2D and doesn't actually work.

Apparently this wiki page is a good reference for this stuff:
https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

Right now I've implemented a one-pass formula that actually requires three
passes, so it's dumb.


keflavich:
@astrofrog, @ChrisBeaumont - this is ready for final review.

PR #226, opened 2015-08-02
Provide a way to disable memory mapping

keflavich:
At the moment, ``SpectralCube.read(..., memmap=False)`` doesn't work.


keflavich:
What do you think of automatically selecting memmap or not (by default, with optional override) for cubes on some size threshold - say, 100MB? (since ~100MB should read nearly instantly from disk, but larger could take significant time even if it fits in memory)

keflavich:
closed by #230.  Not entirely sure why this didn't happen automatically, but I suspect it's because of a rebase (which I think was unnecessary but github forced)

PR #228, opened 2015-08-09
Mask quickview

e-koch:
This stemmed from a discussion in [signal-id](https://github.com/radio-astro-tools/signal-id/issues/31) with @keflavich. ```RadioMask``` in signal-id is intended for interactive mask manipulation and would benefit from the ability to quickly look at slices in the mask.

The changes are working, but I need to add some tests and a way to catch views that will not return a 2D slice (though this will be caught in ```Projection```).


e-koch:
Added a small test.

@keflavich @astrofrog - Would someone mind reviewing?

keflavich:
:+1:.  Ready to merge?

e-koch:
Yup!

PR #230, opened 2015-08-09
Make SpectralCube operate more like a numpy array in some ways

keflavich:
This PR does two things:

 1. Provide the 'base' of self so that things like joblib can operate on
 spectral cubes, understanding that they are memmaps. 
 2. For a related reason, lower dimensional slices should return _data[view],
 not filled_data[view], BUT this doesn't provide the right functionality so we
 want lower dimensional objects to implement their own cubelike masking

The TODO list:

 * [ ] Implement masking for lower-dimensional objects
 * [x] Test the base

### DIFF
--- a/spectral_cube/io/fits.py
+++ b/spectral_cube/io/fits.py
@@ -96,7 +96,7 @@ def read_data_fits(input, hdu=None, **kwargs):
     return array_hdu.data, array_hdu.header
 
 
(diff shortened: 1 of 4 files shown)


keflavich:
assuming this passes, I'll merge it, and the lower-dimensional masking will get put off to a different issue

PR #233, opened 2015-08-11
Use an instance-level cache instead of a global cache, otherwise the …

### LINKED ISSUE #232: Memory leak when computing moment maps
The following example causes a memory increase from one iteration to the next:

```python
import numpy as np
from astropy.io import fits

for i in range(1000):
    
    hdulist = fits.open('l1448_13co.fits', memmap=False)
    sc = SpectralCube.read(hdulist[0])
    hdulist.close()
    
    sc.moment0()
```

Result:

mem

If I remove the ``moment0`` line, everything works fine.

astrofrog:
…cache dictionary contains a reference to all spectral cubes created (in the keys) resulting in a memory leak when spectral cubes go out of scope.

Fixes #232

@ChrisBeaumont @keflavich - does this seem like a sensible solution to you? It works:

mem

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -52,13 +52,13 @@ def cached(func):
     """
     Decorator to cache function calls
     """
-    cache = {}
 
     @wraps(func)
-    def wrapper(*args):
-        if args not in cache:
-            cache[args] = func(*args)
-        return cache[args]
+    def wrapper(self, *args):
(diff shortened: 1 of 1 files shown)


keflavich:
Yes, this looks good.   I had been seeing memory leaks on a large cube iteration that I thought were from memmap, but this looks more plausible.  I don't think we need to add tests, so I'll merge this later today assuming there are no further additions or comments.

PR #234, opened 2015-08-12
Set WCS each time a new WCS is created

keflavich:
set the WCS in masks when they are created to avoid mismatches between the
cube and mask WCSes.  Otherwise, mask wcs != cube wcs errors can arise

### DIFF
--- a/spectral_cube/masks.py
+++ b/spectral_cube/masks.py
@@ -200,6 +200,7 @@ class MaskBase(object):
 
         newwcs = convert_spectral_axis(self._wcs, unit, out_ctype,
                                        rest_value=rest_value)
+        newwcs.wcs.set()
 
         return newwcs
 
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -400,20 +400,22 @@ class SpectralCube(object):
                 # if operation is over two spatial dims
                 if set(axis) == set((1,2)):
                     new_wcs = self._wcs.sub([wcs.WCSSUB_SPECTRAL])
+                    header = self._nowcs_header
                     return OneDSpectrum(value=out,
                                         wcs=new_wcs,
                                         copy=False,
                                         unit=unit,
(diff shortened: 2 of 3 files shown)


keflavich:
I'd like a regression test for this, but I can't reproduce the error that triggered it with the ADV cubes.  If I encounter it again, I'll try to make a test - I don't think this should be merged before then.

keflavich:
OK, my 'regression test' is this:
```
from paths import dpath
from astropy import units as u

cube = SpectralCube.read(dpath('W51Ku_BD_h2co_v30to90_briggs0_contsub.image.fits')).with_spectral_unit(u.km/u.s, velocity_convention='radio')
cube[60]
```
which, without `set_wcs`, reproducibly results in an abort trap.

Running with gdb, I get a *lot* of unrelated warnings, then this:

```
gdb -ex r --args python mwe.py

warning: Could not open OSO archive file "-------src-dir--------/scipy-0.16.0/build/temp.macosx-10.5-x86_64-2.7/libodepack.a"
warning: Could not open OSO archive file "-------src-dir--------/scipy-0.16.0/libgfortran.a"
warning: `-------src-dir--------/scipy-0.16.0/build/temp.macosx-10.5-x86_64-2.7/scipy/ndimage/src/_ni_label.o': can't open to read symbols: No such file or directory.

Program received signal SIGABRT, Aborted.
0x00007fff845fc866 in __pthread_kill () from /usr/lib/system/libsystem_kernel.dylib

(gdb) bt
#0  0x00007fff845fc866 in __pthread_kill () from /usr/lib/system/libsystem_kernel.dylib
#1  0x00007fff853d435c in asn1_TicketFlags_table_units () from /usr/lib/system/libsystem_pthread.dylib
#2  0x0000000000000000 in ?? ()
```

keflavich:
Unfortunately, the tests that produce failures are unreproducible in general, and even if they worked, they would crash *python* and not provide useful tests.  So I'm going to merge.

PR #236, opened 2015-08-17
Add beam info onto the header, if it is defined

e-koch:
Use [Beam.attach_to_header](https://github.com/radio-astro-tools/radio_beam/pull/18) to add the beam information to ```SpectralCube.header``` -- @keflavich 

Missing tests.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -23,6 +23,12 @@ from .io.core import determine_format
 from .ytcube import ytCube
 from .lower_dimensional_structures import Projection, Slice, OneDSpectrum
 
+try:
+    from radio_beam import Beam
+    beam_available = True
+except ImportError:
+    beam_available = False
+
 from distutils.version import StrictVersion
 
 __all__ = ['SpectralCube']
@@ -111,7 +117,7 @@ np2wcs = {2: 0, 1: 1, 0: 2}
 class SpectralCube(object):
 
     def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
-                 header=None, allow_huge_operations=False):
+                 header=None, allow_huge_operations=False, read_beam=True):
 
         # Deal with metadata first because it can affect data reading
         self._meta = meta or {}
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -127,8 +127,7 @@ class SpectralCube(object):
             if self._meta['BUNIT'].lower() == 'jy/beam':
                 self._unit = u.Jy
                 try:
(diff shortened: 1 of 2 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -1982,8 +1982,8 @@ def to_glue(self, name=None, glue_app=None, dataset=None, start_gui=True):
                     return self._glue_app
 
             glue_app.add_datasets(self._glue_app.data_collection, result)
-        
-
```
keflavich: I think my syntax checker misses these... don't know why.

keflavich:
:+1: and :+1: to tests too!

keflavich:
....why does coveralls think the coverage has increased?

astrofrog:
¯\_(ツ)_/¯

keflavich:
your emoticon is not wearing coveralls.

e-koch:
@keflavich - Why is a ```Beam``` object only setup when the units are Jy/beam?

keflavich:
@e-koch good question.  It is only strictly necessary when units are Jy/beam, but it would be nice to have if it's present.  It can be slow, though - CASA FITS headers are 100's of lines of history that can take a human-noticeable amount of time to parse.  Still, maybe we should always try to read a beam?

e-koch:
@keflavich - How about a ```read_beam=True``` like argument? And perhaps it should be disabled by default with a helpful error message for ```spectral_cube.beam``` ?

keflavich:
Actually, I'm ok with it being on by default, but I like the `read_beam` boolean flag.  We need to have it both in the `SpectralCube` init (since that's where beam parsing happens now) and in the `read` method in `io`.

e-koch:
@keflavich @astrofrog - Ready for a review!

keflavich:
:+1: and looks like [tests ran](https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/76150321), so I'll merge

jpinedaf:
There is a bug generated by this implementation in the `write` function, it makes `BMIN=BMAJ`

```
In [39]: from spectral_cube import SpectralCube
In [40]: cube=SpectralCube.read('L1451mm/L1451MM-11_cvel_clean_rob05.fits')
In [41]: cube.beam
Out[41]: Beam: BMAJ=3.08794236183 arcsec BMIN=2.61326932907 arcsec BPA=-19.7857666016 deg
In [42]: cube.write('test.fits')
In [44]: cube2=SpectralCube.read('test.fits')

In [45]: cube2.beam
Out[45]: Beam: BMAJ=3.08794236183 arcsec BMIN=3.08794236183 arcsec BPA=-19.7857666016 deg
```

PR #243, opened 2015-09-21
Make sure median preserves units

keflavich:
There was a bug in which the median did not preserve units.  It should have
caused test failures too since `np.median` was being used instead of
`np.nanmedian`, but I have not observed any such failures, which is worrisome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 language: python
 
 python:
-    - 2.6
     - 2.7
-    - 3.3
     - 3.4
+    - 3.5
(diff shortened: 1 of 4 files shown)


keflavich:
@astrofrog can you help with this?  https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/91541188#L661  I don't understand the error, but apparently test collection is failing on python3.5 + numpy1.9.

keflavich:
@astrofrog if you don't have time to help with this, can you suggest anyone who might know what the problem is?

PR #244, opened 2015-09-28
Add an example document - now with 2 examples!

keflavich:
@astrofrog example 2 in particular is for you.  It may represent a use case
that should be supported within aplpy (or wcsaxes).

These examples really need tests.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -2,6 +2,10 @@
 
 Examples
 ========
+Note that these examples are not tested by continuous integration tests; it is
+possible they will become out-of-date over time.  If you notice any mistakes or
+inconsistencies, please post them at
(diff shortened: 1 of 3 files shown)


astrofrog:
Since the Travis config has now changed, it would make sense to rebase this when you work on fixing the sphinx failure.

docs/examples.rst
```diff
@@ -0,0 +1,178 @@
+.. doctest-skip-all
+
+Examples
+========
+Note that these examples are not tested by continuous integration tests; it is
```
astrofrog: @keflavich - just out of curiosity, why not? Can't we put the cube on data.astropy.org?

keflavich: Maybe?  I don't have any experience doing that.  Let's make CI-testing the examples an Issue.

PR #250, opened 2015-11-04
Preserve metadata when making projections

keflavich:
metadata should be passed to projections so that information like the beam
shape is preserved


e-koch:
I encountered the same build error yesterday. Miniconda is now installed into ```miniconda2```??

```
$ ./miniconda.sh -b
PREFIX=/home/travis/miniconda2
```

PR #253, opened 2015-11-16
Fix pyds9 import (somtimes)

keflavich:
On my system, I have pyds9 installed as ds9, but that seems to be nonstandard.  I don't fully understand the issue, but trying to import both seems to work.

Note that pyds9 is not python3-compatible!

Any recommendations on how to test this?  I guess we'd need to install ds9 and
then ensure that it has the cube in memory.  Probably doable, but I don't have
time now.


keflavich:
we have no real tests for ds9 compatibility, but I'll add those if I find a good way to do so

PR #257, opened 2015-12-23
Strip whitespace from BUNIT

e-koch:
All white-space is now stripped from the BUNIT. I have a header written by CASA that has extra whitespaces in the BUNIT: ```Jy / beam ```.

### DIFF
--- a/spectral_cube/tests/test_spectral_cube.py
+++ b/spectral_cube/tests/test_spectral_cube.py
@@ -1023,6 +1023,7 @@ def test_jybeam_lower():
         np.testing.assert_almost_equal(cube.beam.sr.value,
                                        (((1*u.arcsec/np.sqrt(8*np.log(2)))**2).to(u.sr)*2*np.pi).value)
 
+# Regression test for #257 (https://github.com/radio-astro-tools/spectral-cube/pull/257)
 def test_jybeam_whitespace():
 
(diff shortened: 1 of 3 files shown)


keflavich:
Darn, I was worried about this corner case but I thought FITS automatically stripped the whitespace.

Might it be better to use `re`?  e.g., `re.sub("\s","",self._meta['BUNIT'].lower())`?

keflavich:
Also, could you add a test for this?

e-koch:
Switched to ```re```. It does look a bit cleaner.

@keflavich, mind having a second look?

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -1022,7 +1022,17 @@ def test_jybeam_lower():
         assert hasattr(cube, 'beam')
         np.testing.assert_almost_equal(cube.beam.sr.value,
                                        (((1*u.arcsec/np.sqrt(8*np.log(2)))**2).to(u.sr)*2*np.pi).value)
-        
+
+def test_jybeam_whitespace():
+
```
keflavich: Add in a comment here that this is a regression test for #257

keflavich:
:+1: I'll merge it with the comment added.

keflavich:
or you can merge it

PR #256, opened 2015-12-30
WCS information lost in unit conversion of LowerDimensionalObject

keflavich:
A similar custom ```to``` attribute should be defined (as is already done for ```SpectralCube```).

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -20,6 +20,10 @@ class LowerDimensionalObject(u.Quantity):
         return self._meta
 
     @property
+    def mask(self):
+        return self._mask
+
+    @property
     def header(self):
         header = self._header
         # This inplace update is OK; it's not bad to overwrite WCS in this
@@ -67,6 +71,24 @@ class LowerDimensionalObject(u.Quantity):
         else:
             raise ValueError("Unknown format '{0}' - the only available "
                              "format at this time is 'fits'")
+
+    def to(self, unit, equivalencies=[]):
+        """
+        Return a new ``LowerDimensionalObject'' of the same class with the
+        specified unit.  See `astropy.units.Quantity.to` for furthe details.
+        """
+        converted_array = u.Quantity.to(self, unit,
+                                        equivalencies=equivalencies).value
+
+        # use private versions of variables, not the generated property
+        # versions
+        new = super(self, LowerDimensionalObject).__new__(
+            value=converted_array, unit=unit, copy=True,
+            wcs=self._wcs, meta=self._meta, mask=self._mask,
+            header=self._header)
+
+        return new
+
 
 class Projection(LowerDimensionalObject):
 
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -99,12 +99,17 @@ class LowerDimensionalObject(u.Quantity):
 
         if new_qty.ndim < 2:
             # do not return a projection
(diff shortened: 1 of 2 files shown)


keflavich:
@e-koch care to review this?

spectral_cube/lower_dimensional_structures.py
```diff
@@ -68,6 +72,52 @@ def write(self, filename, format=None, overwrite=False):
...
+    def __getitem__(self, key):
+        """
+        Return a new ``LowerDimensionalObject'' of the same class while keeping
+        other properties fixed.
+        """
+        new_qty = super(LowerDimensionalObject, self).__getitem__(key)
+
+        if new_qty.ndim < 2:
+            # do not return a projection
+            return u.Quantity(new_qty)
```
e-koch: I'm not sure what should be returned here. If a `Quantity` is returned, then you still lose all of the WCS info. If you have a PV slice, you would want a `OneDSpectrum` when slicing along the spectral axis, but something else when slicing in the spatial dimension.

Could we check for the axis type when a slice reduces `ndim<2`? A `OneDSpectrum` would be returned when the remaining axis is spectral. I'm not sure how to treat 1D spatial slices. Treat it like a mock-projection with a 1 pixel second dimension or define a spatial 1D slice class? Is there a use case for such a thing?

e-koch:
@keflavich - Apart from the one comment, it looks good! I tested on the use case the brought this up  and everything checks out.

Depending on what you think about the comment above, `test_slices_of_projections_not_projections` may need to be altered/become more specific on the returned object.

spectral_cube/lower_dimensional_structures.py
keflavich: There might be.  I chose to ignore it for now because it's not trivial, but I think we could inspect the WCS and see if one of the axes is spectral; if so, we'd then return the `OneDSpectrum`

keflavich: But I think for single pixels, we really want to drop all the way to a `Quantity` with no wcs information

keflavich:
There is a bug in WCS that prevents creating 1D slices from 2D WCSs that are half celestial, half spectral:

```
In [15]: w2 = cube[:,0,:].wcs
WARNING: FITSFixedWarning: 'spcfix' made the change 'Changed CTYPE3 from 'VELO-LSR' to 'VOPT', and SPECSYS to 'LSRK''. [astropy.wcs.wcs]

In [16]: w2.sub([wcs.WCSSUB_SPECTRAL])
  File "<ipython-input-16-0b9117b7127a>", line 1, in <module>
    w2.sub([wcs.WCSSUB_SPECTRAL])
  File "/Users/adam/repos/astropy/astropy/wcs/wcs.py", line 541, in sub
    copy = self.deepcopy()
  File "/Users/adam/repos/astropy/astropy/wcs/wcs.py", line 538, in deepcopy
    return copy.deepcopy(self)
  File "/Users/adam/anaconda/envs/astropy35/lib/python3.5/copy.py", line 166, in deepcopy
    y = copier(memo)
  File "/Users/adam/repos/astropy/astropy/wcs/wcs.py", line 514, in __deepcopy__
    copy.deepcopy(self.wcs, memo),
  File "/Users/adam/anaconda/envs/astropy35/lib/python3.5/copy.py", line 166, in deepcopy
    y = copier(memo)
InconsistentAxisTypesError: ERROR 4 in wcs_types() at line 2516 of file cextern/wcslib/C/wcs.c:
Unmatched celestial axes.
```

I'll make a separate issue for this.

PR #258, opened 2016-01-05
Updated .travis.yml to use ci-helpers

astrofrog:


astrofrog:
Still investigating memory issues - I think there is a memory leak in astropy.wcs

astrofrog:
@keflavich - this is ready for review!

astrofrog:
Note that some of the configurations still take 10-20 minutes to run (as was the case prior to this PR), which is what led me to investigate this memory leak in Astropy: https://github.com/astropy/astropy/issues/4441 - but I just realized that actually the builds with Astropy dev are already a lot faster, so I think this is just an issue with the Astropy stable release. I'm not 100% sure what was fixed since 1.1 that would cause this though. Anyway, I'll look into it separately.

spectral_cube/cube_utils.py
```diff
@@ -191,8 +191,13 @@ def __iter__(self):
         raise Exception("You need to specify a slice (e.g. ``[:]`` or "
                         "``[0,:,:]`` in order to access this property.")
 
-def is_huge(cube, threshold=1e8):
-    if cube.size < threshold:  # smallish
+
+# TODO: make this into a proper configuration item
+# TODO: make threshold depend on memory?
```
keflavich: I think it's reasonable to *not* make the threshold depend on memory, but make this a configuration item.  Even with lots of memory, operations on huge cubes can take forever, and I think it's better to be up front about that.

keflavich: ...oh, the note was from me.  Well, I can argue with myself!

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -74,18 +75,31 @@ def test_huge_disallowed():
...
     cube, data = cube_and_raw('vda_Jybeam_lower.fits')
 
-    data = np.empty([1e2,1e3,1e3])
     cube = SpectralCube(data=data, wcs=cube.wcs)
 
-    assert cube._is_huge
+    assert not cube._is_huge
+    
+    # We need to reduce the memory threshold rather than use a large cube to
+    # make sure we don't use too much memory during testing.
```
keflavich: I think we still need some genuinely large cube tests, though perhaps not to be run regularly on travis... that may be a different issue, though.  See, e.g., #259, which was uncaught because we don't have automated tests on gigantocubes

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -602,7 +617,7 @@ def test_yt(self):
         ds2.quan(1.0,unit2)
         ds3.quan(1.0,unit3)
 
-    @pytest.mark.skipif(yt_version < StrictVersion('3.0.1'), reason='yt 3.0 has a FITS-related bug')
+    @pytest.mark.skipif('YT_LT_301')
```
keflavich: in this case, I think it is important to keep the `reason` string

keflavich:
Looks great.  I only have one comment that warrants change (the `reason` in `skipif`), everything else is ready to merge and I was just initiating conversation.

spectral_cube/cube_utils.py
astrofrog: I've opened an issue for the config item: https://github.com/radio-astro-tools/spectral-cube/issues/261

spectral_cube/tests/test_spectral_cube.py
astrofrog: Sounds good - I've added an issue here: https://github.com/radio-astro-tools/spectral-cube/issues/262 - maybe we want a ``--large-cube-tests`` option to enable those tests

spectral_cube/tests/test_spectral_cube.py
astrofrog: Fixed

astrofrog:
@keflavich - I addressed your comment and Travis is still passing (and I think the coverage failure is just due to random fluctuations) so I'll go ahead and merge this.

PR #259, opened 2016-01-06
Memory Error in SpectralCube._pix_size

keflavich:
This arose when computing the 0th moment of a cube.
Running `cube._pix_cen()` gives the following:
```
MemoryError                               Traceback (most recent call last)
<ipython-input-2-a118f0eff323> in <module>()
----> 1 cube._pix_size()

/home/eric/Dropbox/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in wrapper(self, *args)
     61         # The cache lives in the instance so that it gets garbage collected
     62         if func not in self._cache:
---> 63             self._cache[func] = func(self, *args)
     64         return self._cache[func]
     65 

/home/eric/Dropbox/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in _pix_size(self)
   1345 
   1346         # Take spectral units into account
-> 1347         dspectral = dspectral * self._spectral_scale
   1348 
   1349         return dspectral, dy, dx

MemoryError: 
```

It's a fairly large cube (~30 Gb). I read it in with:
```
cube = SpectralCube.read("M33_14B-088_HI.clean.image.fits", mode='denywrite', memmap=True)
```

Commenting out line 1347 does seems to fix the issue, and the units of the moment map remain correct. The `cube._spectral_scale` is a float, not a quantity. Setting it to a quantity (e.g., km/s) still gives the memory error.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -1271,15 +1271,32 @@ class SpectralCube(object):
 
         return spectral, y, x
 
+    @cached
     def _pix_size_slice(self, axis):
         """
         Return the size of each pixel along any given direction.  Assumes
-        pixels have equal size.
+        pixels have equal size.  Also assumes that the spectral and spatial
+        directions are separable, which is enforced throughout this code.
+
+        Parameters
+        ----------
+        axis : 0, 1, or 2
+            The axis along which to compute the pixel size
+
+        Returns
+        -------
+        Pixel size in units of either degrees or the appropriate spectral unit
         """
         if axis == 0:
-            return self.wcs.wcs.cdelt[0] * self._spectral_scale
+            # note that self._spectral_scale is required here because wcs
+            # forces into units of m, m/s, or Hz
+            return self.wcs.pixel_scale_matrix[2,2] * self._spectral_scale
+        elif axis in (1,2):
+            # the pixel size is a projection.  I think the pixel_scale_matrix
+            # must be symmetric, such that psm[axis,:]**2 == psm[:,axis]**2
+            return np.sum(self.wcs.pixel_scale_matrix[axis,:]**2)**0.5
         else:
-            return self.wcs.wcs.cdelt[axis]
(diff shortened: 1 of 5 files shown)


keflavich:
Actually, I think this might be a bad interaction between caching and `_pix_cen`/`_pix_size`: caching something that is supposed to be cheap to compute can be bad.

keflavich:
I think I caught a few problems, including unintentional massive duplication of data because of an order of operations issue and some unnecessary code.  I think the core of this problem, though, was something like this:
```
# small array
x = np.arange(1000)
# small array (same values in memory)
y = np.broadcast_arrays(x[:,None,None], x[None,:,None], x[None,None,:])
# now, the full 10^9 floats must be allocated!  Huge array!
y = y - 1
```
which was easily avoided by moving the subtraction before the broadcasting

keflavich:
@e-koch I think this solves it now

spectral_cube/spectral_cube.py
```diff
@@ -1264,14 +1265,38 @@ def _pix_cen(self):
...
+            The axis along which to compute the pixel size
+
+        Returns
+        -------
+        Pixel size in units of either degrees or the appropriate spectral unit
+        """
+        if axis == 0:
+            # note that self._spectral_scale is required here because wcs
+            # forces into units of m, m/s, or Hz
+            return self.wcs.pixel_scale_matrix[2,2] * self._spectral_scale
```
e-koch: @keflavich - I get a sign flip since the cube I'm testing with locally has `self.wcs.pixel_scale_matrix[2,2] < 0`. Is it safe to force an absolute value here? Or WCS isn't doing something consistently since `_, _, spectral = self._wcs.all_pix2world(xpix, ypix, zpix, 0)` returns positive values in `_pix_size`.

keflavich: Yes, absolute value is the right thing to do here.  I forgot we're asking for absolute scale.

e-koch:
:+1:

PR #263, opened 2016-01-07
Don’t use 2to3, and fix failing test

astrofrog:


astrofrog:
This is straightforward, so I'll go ahead and merge. I think we must have had a 2/3 compatible code base at some point, and accidentally re-introduced the use_2to3 flag.

keflavich:
I think that's correct, though perhaps we never unset the 2to3 flag, which I think package_template had on by default.  Anyway, :+1:

PR #267, opened 2016-01-20
Fix capitalization of Jy/beam in warning

astrofrog:
Otherwise it made me think the unit was JY/BEAM in the header and it wasn't recognized because that isn't valid FITS. But the capitalization was correct and it's just that I didn't have radio_beam installed.


astrofrog:
@keflavich - looks ok?

PR #268, opened 2016-01-20
More convenient beam area determination

keflavich:
add a cube utility attribute to get the area of the beam in pixels when a beam
is defined

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -254,13 +254,11 @@ class SpectralCube(object):
         try:
             self.beam = Beam.from_fits_header(header)
             self._meta['beam'] = self.beam
-            self.beam_area_pixels = self._beam_area_pixels
+            self.beam_area_pixels = (self.beam.sr /
+                                     (wcs.utils.proj_plane_pixel_area(self.wcs) *
+                                      u.deg**2)).to(u.dimensionless_unscaled).value
         except:
             warnings.warn("Could not parse beam information from header.")
-
-    @property
-    def _beam_area_pixels(self):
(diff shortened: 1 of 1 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -254,10 +254,15 @@ def _try_load_beam(self, header):
         try:
             self.beam = Beam.from_fits_header(header)
             self._meta['beam'] = self.beam
+            self.beam_area_pixels = self._beam_area_pixels
         except:
             warnings.warn("Could not parse beam information from header.")
 
     @property
+    def _beam_area_pixels(self):
+        return (self.beam.sr/(wcs.utils.proj_plane_pixel_area(self.wcs)*u.deg**2)).to(u.dimensionless_unscaled).value
```
astrofrog: Do we need a property for this, or can we just do the calculation in-line on line 257?

keflavich: my intent was to make this an immutable value, but it probably still is not because it's going to grab the value from the property... so, no, the property is pointless

astrofrog:
Would ``pixels_per_beam`` read better as a name for the value? (I think it would in equations)

e-koch:
@keflavich - I think there's a conversion missing here: `self.beam.sr.to(u.deg**2)`. The try/except is masking a `UnitConversionError`.

keflavich:
@e-koch : first, good point that we should not catch `UnitConversionError`s.  However, `sr/deg**2` should be equivalent to `dimensionless_unscaled`.

e-koch:
@keflavich - Ah! I was just assuming it was a units issue without actually checking... 

It's an `AttributeError` I'm getting when calculating `pixels_per_beam`, since `self._wcs` hasn't been set yet. Are you getting the "Could not parse beam information from header." every time you read in a cube?

PR #270, opened 2016-01-25
Set _data, _wcs earlier in init

e-koch:
Set `_data` and `_wcs` earlier in `__init__`. From #268, `_wcs` needs to be set when reading the beam info in `_try_load_beam`. It seems more appropriate to do things in this order as well, since technically `header=None` would be passed into `_try_load_beam`, instead of being set to an empty header.

@keflavich


keflavich:
I missed this, but it is (nearly) identical to #271.

e-koch:
Cool. I'll close this one.

keflavich:
I've merged this into #271

PR #274, opened 2016-01-27
Return spectral axis units in OneDSpectrum

e-koch:
No units were being returned with the spectral axis in OneDSpectrum.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+0.4.0 (unreleased)
+------------------
+
+ - added new linewidth_sigma and linewidth_fwhm methods to SpectralCube for
+   computing linewidth maps, and make sure the documentation is clear that
+   moment(order=2) is a variance map.
+   (https://github.com/radio-astro-tools/spectral-cube/pull/275)
+
 0.3.1 (2016-02-04)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,27 @@
+0.3.1 (2016-02-04)
+------------------
+
+ - Preserve metadata when making projections
+   (https://github.com/radio-astro-tools/spectral-cube/pull/250)
+ - bugfix: cube._data cannot be a quantity
+   (https://github.com/radio-astro-tools/spectral-cube/pull/251)
+ - partial fix for ds9 import bug
+   (https://github.com/radio-astro-tools/spectral-cube/pull/253)
+ - preserve WCS information in projections
+   (https://github.com/radio-astro-tools/spectral-cube/pull/256)
+ - whitespace stripped from BUNIT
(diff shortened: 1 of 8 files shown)


keflavich:
:+1:, but we should also include conversion to the `OneDSpectrum`'s unit (as we do with `SpectralCube`)

keflavich:
e.g., https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_cube.py#L1853

e-koch:
@keflavich  - So `__new__` would accept a `spectral_unit` argument?

e-koch:
I suppose a `with_spectral_unit` method should also be added.

e-koch:
Adding `with_spectral_unit` isn't as trivial as I had hoped. The new WCS from `spectral_axis.convert_spectral_axis` reverts to the original units when `newwcs.wcs.set()` is called. @keflavich - Any idea why that is happening?

keflavich:
@e-koch yes.  WCS has a method... forget the name, but something like "fixunits"... that is called every time any manipulation is made of the WCS, and it automatically converts to a specific set of units (deg, deg, [m/s, hz, m]).  This is why we change the units on outputting them in https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_cube.py#L1853 rather than changing the underlying WCS.  So, any tracking of units has to happen at the class level, not at the WCS level.  This is a fundamental limitation of the wcslib library, so it's not something that can be worked around even in astropy.

keflavich:
This looks pretty good to me now, but there is a lot of duplicated code between `OneDSpectrum` and `SpectralCube` now (only one function, but a lot of lines).  Is this a case where we want some parent class to inherit the function from?  Is that called a mixin class?  @astrofrog, you might know the terminology better than I do.

e-koch:
I think this is going to need a `_new_cube_with` to method as well (or add this to `with_spectral_unit`). The WCS unit conversion is still an issue here. I'm probably going to be delayed with adding changes until ~ Monday.

I think a parent class with common functionality would be a good idea. In another branch, I've added `world` and `spatial_coordinate_map` functionality to `Projection` so you can extract regions based on world coordinates (like `subcube`). The vast majority of the additions are duplicates from `SpectralCube`.

We could split `with_spectral_unit` and maybe `spectral_axis` functionality into its own class, and a generalized version of `world`, `spatial_coordinate_map`, `world_extrema`, etc... into another. Maybe?

keflavich:
@e-koch that sounds like a plan, and perhaps a good idea for a half-day sprint next week

keflavich:
merged manually with a rebase... ugly weird history here; I tried PRing onto e-koch's branch then rebased onto master and that lead to a royal mess, so I just dealt with it offline

PR #277, opened 2016-02-12
Tests for lower dimensional objects

e-koch:
Testing for:

* [x] `with_spectral_unit` in `OneDSpectrum`

### DIFF
--- a/spectral_cube/tests/test_projection.py
+++ b/spectral_cube/tests/test_projection.py
@@ -131,6 +131,7 @@ def test_onedspectrum_specaxis_units():
 
     assert p.spectral_axis.unit == u.Unit("m/s")
 
+
 def test_onedspectrum_with_spectral_unit():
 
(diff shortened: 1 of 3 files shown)


spectral_cube/tests/test_projection.py
```diff
@@ -130,3 +130,13 @@ def test_onedspectrum_specaxis_units():
...
+
+def test_onedspectrum_with_spectral_unit():
+
+    test_wcs = WCS(naxis=1)
+    test_wcs.wcs.cunit = ["m/s"]
+
+    p = OneDSpectrum(twelve_qty_1d, wcs=test_wcs)
+    p_new = p.with_spectral_unit(u.km/u.s)
+
+    assert p_new.spectral_axis.unit == u.Unit("km/s")
```
keflavich: can we add a check that the values are right?

PR #287, opened 2016-03-16
added example showing how to loop through regions in a region file

aakepley:
I've added an example showing how loop over a set of regions in a region file and extract sub_cubes for each region.


keflavich:
Thanks, looks good!

PR #288, opened 2016-03-16
Bugfix: spectral unit initialization in with_spectral_unit

keflavich:
fix a bug: if you create a new spectral cube with stringy units ('km/s'), it
will fail on writing because other functions assume _spectral_unit is a u.Unit


keflavich:
I don't know why the travis build never triggered, but I'm going to merge this.

PR #291, opened 2016-04-24
Broadcast np.ndarray masks in with_mask

e-koch:
Set the shape of the `BooleanArrayMask` to the shape of the data so the `np.ndarray` mask is broadcasted. Without setting `shape`, the mask keeps its original shape, which may not be the shape of the data.

@keflavich is the `is_broadcastable_and_smaller` check still needed here? I think a check is already done within `BooleanArrayMask`.


keflavich:
@e-koch I think you're right that there are redundant `is_broadcastable` tests, but I believe *this* is the most correct one because it checks before trying to create the mask, thus avoiding possible overheads in mask creation and combination (though those are small).  I am not entirely sure both are necessary though.

e-koch:
@keflavich Alright. Either way, it's a minuscule amount of redundancy.

PR #289, opened 2016-03-26
Add a cube reprojection method

keflavich:
cc @cdeil, @adonath, @astrofrog

This is a long-awaited feature, and all the pieces seem to finally be in place
to have it.  It won't work for cubes with different spectral axes until
https://github.com/astrofrog/reproject/pull/96 is merged.

TODO:

 * [ ] add ``reproject`` as a soft dependency
 * [ ] Increase efficiency of reprojection by pre-slicing the cube to contain
   the minimal subcube represented by the reprojection target header

PR #292, opened 2016-04-29
Support varying spatial resoln along the spectral axis

keflavich:
Infrastructure for handling varying resolution from new CASA tclean cubes

cc @low-sky, @e-koch, @astrofrog

See related https://github.com/radio-astro-tools/radio_beam/pull/35

### DIFF
--- a/spectral_cube/cube_utils.py
+++ b/spectral_cube/cube_utils.py
@@ -4,6 +4,8 @@ import numpy as np
 from astropy.wcs import (WCSSUB_SPECTRAL, WCSSUB_LONGITUDE, WCSSUB_LATITUDE)
 from . import wcs_utils
 from astropy import log
+from astropy.io.fits import BinTableHDU
+from astropy import units as u
 
 def _fix_spectral(wcs):
     """
@@ -230,3 +232,24 @@ def iterator_strategy(cube, axis=None):
     if cube.size < 1e8:  # smallish
         return 'cube'
     return 'slice'
+
+
+def beams_to_bintable(beams):
+    """
+    Convert a list of beams to a CASA-style BinTableHDU
+    """
+    metakeys = beams[0].meta.keys()
+    dtypes = ([('BMAJ','float'),
+               ('BMIN','float'),
+               ('BPA','float'),] +
+              [(key, type(val)) for key,val in beams[0].meta.items()]
+             )
+    beam_table = np.recarray(len(beams), dtype=dtypes)
+    beam_table['BMAJ'] = [bm.major.to(u.arcsec).value for bm in beams]
+    beam_table['BMIN'] = [bm.minor.to(u.arcsec).value for bm in beams]
+    beam_table['BPA'] = [bm.pa.to(u.deg).value for bm in beams]
+    for key in metakeys:
+        beam_table[key] = [bm.meta[key] for bm in beams]
+
+    bmhdu = BinTableHDU(beam_table)
+    return bmhdu
--- a/spectral_cube/cube_utils.py
+++ b/spectral_cube/cube_utils.py
@@ -253,3 +253,21 @@ def beams_to_bintable(beams):
 
     bmhdu = BinTableHDU(beam_table)
     return bmhdu
+
+def average_beams(beams):
+    """
+    Average the beam major, minor, and PA attributes.
+
+    This is usually a dumb thing to do!
+    """
+
+    from radio_beam import Beam
+    from astropy.stats import circmean
+
+    major = u.Quantity([bm.major for bm in beams], u.deg)
+    minor = u.Quantity([bm.minor for bm in beams], u.deg)
+    pa = u.Quantity([bm.pa for bm in beams], u.deg)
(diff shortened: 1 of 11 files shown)


e-koch:
:+1: for multi-beam tests

spectral_cube/spectral_cube.py
```diff
@@ -2287,6 +2290,203 @@ def find_lines(self, velocity_offset=None, velocity_convention=None,
 
         return result
 
+    def reproject(self, header, order='bilinear'):
```
e-koch: `@warn_slow` or a warning for potentially needing a lot of memory

keflavich: Yep, but this belongs on https://github.com/radio-astro-tools/spectral-cube/pull/289

keflavich:
Yay, all passing!  I think this is ready to merge, though since it's big I'd like a review from @astrofrog or @e-koch.  I might wait a few days at least and do some local testing (real use) to see how it fares.

keflavich:
I think we now have an interdependency issue for the tests between this and `radio_beam`

keflavich:
The latest error has something to do with travis reconfiguration; apparently the wrong numpy version is being run or something like that?

keflavich:
One of the failures was that conda doesn't have bottleneck for python3.5, only python3.4.  I don't know why...

keflavich:
yay, improved coverage!  But, I can't reproduce these numpy 1.9 errors locally.

spectral_cube/spectral_cube.py
```diff
@@ -2287,6 +2299,363 @@ def find_lines(self, velocity_offset=None, velocity_convention=None,
...
+            if maxdiff > th:
+                errormessage += ("Beam {2}s differ by up to {0}x, which is greater"
+                                 " than the threshold {1}\n".format(maxdiff,
+                                                                    threshold,
+                                                                    qtyname
+                                                                   ))
+        if errormessage != "":
+            raise ValueError(errormessage)
+
+    def _average_beams(self, threshold):
```
e-koch: `threshold` isn't being checked in here. The `_check_beam_areas` in `_handle_beam_areas_wrapper` could be moved into here to ensure the threshold check is performed for any beam averaging.

spectral_cube/spectral_cube.py
```diff
@@ -2287,6 +2299,363 @@ def find_lines(self, velocity_offset=None, velocity_convention=None,
...
+            # check that the spectral axis is being operated over
+            # moments are a special case b/c they default to axis=0
+            need_to_handle_beams = ('axis' in kwargs and kwargs['axis']==0
+                                    or (hasattr(kwargs['axis'], '__len__') and
+                                        0 in kwargs['axis'])
+                                    or ('axis' not in kwargs and 'moment' in
+                                        function.__name__))
+
+            if need_to_handle_beams:
+                self._check_beam_areas(beam_threshold)
```
e-koch: See comment above.

spectral_cube/spectral_cube.py
```diff
@@ -2287,6 +2299,363 @@ def find_lines(self, velocity_offset=None, velocity_convention=None,
...
+
+        qtys = dict(area = u.Quantity(self.beams, u.sr),
+                    major = u.Quantity([bm.major for bm in self.beams], u.deg),
+                    minor = u.Quantity([bm.minor for bm in self.beams], u.deg),
+                    pa = u.Quantity([bm.pa for bm in self.beams], u.deg),
+                   )
+
+        errormessage = ""
+
+        for qtyname, qty in qtys.items():
```
e-koch: `cube_utils.average_beams` should be used to get the mean values (for consistency).

For the PA, I guess the comparison should be with the max absolute difference from the mean?

spectral_cube/cube_utils.py
```diff
@@ -230,3 +232,42 @@ def iterator_strategy(cube, axis=None):
...
+    """
+
+    from radio_beam import Beam
+    from astropy.stats import circmean
+
+    major = u.Quantity([bm.major for bm in beams], u.deg)
+    minor = u.Quantity([bm.minor for bm in beams], u.deg)
+    pa = u.Quantity([bm.pa for bm in beams], u.deg)
+    new_beam = Beam(major=major.mean(), minor=minor.mean(),
+                    pa=circmean(pa, weights=major/minor))
```
e-koch: Is there a big difference if you just weight be `major`? The minor axis should be independent from the vector defining the PA. `major/minor` weights the most asymmetric beams the most, and I'm not sure if that's what we want. I guess the most asymmetric beams are where the PA matters the most.

e-koch:
@keflavich - I've added some minor comments. I haven't had a chance to test locally yet (though I'm working on the PR branch and it hasn't broken my normal workflow).

spectral_cube/cube_utils.py
keflavich: That was my thinking.  I don't know what a correct weighting scheme would be; probably needs to be determined

keflavich:
OK, if this passes I'll merge.  But changelog entry, right...

e-koch:
:+1:

PR #298, opened 2016-05-12
Mask out planes with NaN beams

keflavich:


keflavich:
I have a problem here:  Why did this fail:
https://github.com/keflavich/spectral-cube/blob/426400624cf949aea8053ae7232265d1b07c4348/spectral_cube/spectral_cube.py#L2450

but this worked:
https://github.com/keflavich/spectral-cube/blob/master/spectral_cube/tests/test_masks.py#L398

as far as I can tell, they're the same code, but the first instance resulted in deep failures while the second has always worked.

keflavich:
@e-koch @low-sky @astrofrog any of you up for reviewing this?

PR #301, opened 2016-05-25
Handle WCS reversal & fix 1D slice headers

keflavich:
WCS reversal, e.g. cube[::-1,:,:], is allowed now but does *not* work because
of https://github.com/astropy/astropy/pull/4962.  This PR partially implements
a reverse indexing.

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -26,6 +26,9 @@ class LowerDimensionalObject(u.Quantity, BaseNDClass):
         if self.wcs is not None:
             header.update(self.wcs.to_header())
         header['BUNIT'] = self.unit.to_string(format='fits')
+        for keyword in header:
+            if 'NAXIS' in keyword:
+                del header[keyword]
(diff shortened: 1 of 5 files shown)


keflavich:
@astrofrog this is related to https://github.com/astropy/astropy/pull/4962 and it even replicates some of the code there (....I think it was written here first?), but right now it's a quick workaround and not a complete solution.  Interestingly enough, though, I think you could do `mywcs[::-1,:,:][5:10:2,:,:]` and that would now be correct (i.e., what you expect from `mywcs[5:10:-2,:,:]`); not entirely sure though.

astrofrog:
> I think you could do mywcs[::-1,:,:][5:10:2,:,:] and that would now be correct (i.e., what you expect from mywcs[5:10:-2,:,:]); not entirely sure though

Are you sure those are equivalent? If you do [::-1] to reverse then do [5:10], won't it then be equivalent to selecting [0:5]?

keflavich:
Right, no, those would not be equivalent.  But it is possible to achieve `[5:10:-2]` by chaining together some subset of indexing commands... I guess, as you identified, `[::-1][0:5:2]`

PR #303, opened 2016-05-25
Fix call to reproject_interp

e-koch:
`shape_out` has its own kwarg.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2369,7 +2369,8 @@ class SpectralCube(BaseNDClass, SpectralAxisMixinClass):
 
         newcube, newcube_valid = reproject_interp((self.filled_data[:],
                                                    self.header),
-                                                  (newwcs, shape_out),
+                                                  newwcs,
+                                                  shape_out=shape_out,
                                                   order=order,
                                                   independent_celestial_slices=True)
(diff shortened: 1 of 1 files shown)


keflavich:
@e-koch any tests you want to add?  This function is not tested at all now.

e-koch:
@keflavich It couldn't handle the cube reprojection I need to do right now (way too big), so I haven't actually used it yet. I'll add an issue to remind me later.

keflavich:
@e-koch OK.  If you're having trouble with really large cubes, it might be worth trying to re-make the iterator with a write-to-disk at each step.  I've recently had to do this for some files, and I ended up using astropy.io.fits directly instead of spectral-cube.  It would be better if we could make SC work with this.

The approach we'll need to take is to open the file with mode='update' (which I hope can work on arbitrarily large files....) and then edit each frame in turn.  It may require specifying both an input & an output file on disk, which is somewhat contrary to the general approach of spectral-cube, but is necessary for this sort of operation.   Now that I think of it, having a general mechanism for streaming from one file to another seems like a killer app for SC.

e-koch:
Agreed. That's basically what I've been having to do with large manipulations on big cubes. I'm up for adding this in, but it'll be a few weeks before I can get to it.

PR #306, opened 2016-05-26
Reproject install and version check

e-koch:
Checks if `reproject` is installed and if it's version is `<0.3`

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2369,7 +2369,7 @@ class SpectralCube(BaseNDClass, SpectralAxisMixinClass):
         # Need version > 0.2 to work with cubes
         from distutils.version import LooseVersion
         if LooseVersion(version) < "0.3":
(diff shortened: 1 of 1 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -2359,6 +2359,19 @@ def reproject(self, header, order='bilinear'):
...
+            from reproject.version import version
+        except ImportError:
+            raise ImportError("Requires the reproject package to be"
+                              " installed.")
+
+        # Need version > 0.2 to work with cubes
+        from distutils.version import LooseVersion
+        if LooseVersion(version) < "0.3":
+            raise Warning("Requires version >=0.3 of reproject. The current "
+                          "version is: {}".format(version))
```
keflavich: I think python >=2.7 supports "{}".format(...), but <2.7 didn't.... if it fails, this is a place to look (then use "{0}".format() instead)

e-koch: I think we're only testing for >=2.7 now.

PR #316, opened 2016-06-14
Raise exception when attempting to subcube an empty set

keflavich:
Would it be better to return empty?  Please chime in if you have an opinion.


e-koch:
I like returning empty. Then doing operations like looping over subcubes requires an if statement instead of try/excepts everywhere.

spectral_cube/spectral_cube.py
```diff
@@ -1690,6 +1690,10 @@ def subcube_slices_from_mask(self, region_mask, spatial_only=False):
         include = region_mask.include(self._data, self._wcs,
                                       wcs_tolerance=self._wcs_tolerance)
 
+        if not any(include):
```
e-koch: Any advantage to using `include.any()` instead?

PR #328, opened 2016-07-07
broken lower-dimensional-object units

keflavich:
With the latest astropy, lower-dimensional-structures, which inherit from astropy Quantitys, fail.  Example:

```
In [3]: x = OneDSpectrum(np.arange(5), unit=u.km**2/u.s**2)

In [4]: x
Out[4]: <OneDSpectrum [ 0., 1., 2., 3., 4.] km2 / s2>

In [5]: x**0.5
Out[5]: <OneDSpectrum [ 0.        , 1.        , 1.41421356, 1.73205081, 2.        ] km2 / s2>
```

The units aren't being updated.

cc @mhvk: can you offer any insight on what might be happening?  This is probably related to recent changes you made to astropy.quantity, but I'm not sure what exactly.

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -119,11 +119,11 @@ class LowerDimensionalObject(u.Quantity, BaseNDClass):
         return new
 
     def __array_finalize__(self, obj):
-        self._unit = getattr(obj, '_unit', None)
         self._wcs = getattr(obj, '_wcs', None)
         self._meta = getattr(obj, '_meta', None)
         self._mask = getattr(obj, '_mask', None)
         self._header = getattr(obj, '_header', None)
+        super(LowerDimensionalObject, self).__array_finalize__(obj)
 
     @property
     def __array_priority__(self):


keflavich:
@mhvk nevermind, I found the issue.  `_set_unit` replaced a line in `__array_finalize__`

mhvk:
@keflavich - great that you found the difference, but not great that there was a problem in the first place... Your solution is the right one, though: always call `super` on such routines -- I presume that with this change the code also works with older versions of astropy?

keflavich:
@mhvk yes, it does.  This was a case where I wasn't sure whether `super` was appropriate in the first place; it turns out it was, but I wasn't certain of that.  

The current travis failure has to do with an astropy-helpers issue in another one of spectral-cube's dependencies.

keflavich:
The issues are due to configuration issues on pvextractor.  I'm giving the fix one more try, but then will merge this and cut a new release to get past the circuitous interdependency failures (spectral-cube and pvextractor depend on each other for some things, so their tests will both fail until at least one is corrected)

PR #332, opened 2016-07-28
pyregion pixel regions are incorrectly modified

keflavich:
This PR fixes a bug in which a pixel region would be incorrectly modified if passed to `subcube_from_ds9region`


keflavich:
Only #335 issues remain.

PR #334, opened 2016-08-02
Set CONDA_CHANNELS

keflavich:
commit message says all:
add specific conda channels to .travis as per @bispocz's recommendations


keflavich:
The only remaining error is yt-related (https://github.com/radio-astro-tools/spectral-cube/issues/335).  This is safe so I'll merge

PR #346, opened 2016-08-17
DOCS: Instructions on making fake WCS

keflavich:
cc @e-koch @scog1234

Add instructions on how to create the spatial component of a WCS for synthetic images


e-koch:
Can we gracefully handle unit conversions in `CDELT`? If not, there should be a note about altering headers to have pixel units

e-koch:
👍

PR #356, opened 2017-01-05
Handle equal beams when convolving

e-koch:
Added handling for when the beam in `convolveto` is equal to the channel beam.

Also added a function to return the largest beam from a list of beams.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,7 @@
 0.4.0 (2016-09-06)
 ------------------
+ - Handle equal beams when convolving cubes spatially.
+   (https://github.com/radio-astro-tools/spectral-cube/pull/356)
  - Whole cube convolution & reprojection has been added, including tools to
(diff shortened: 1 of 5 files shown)


keflavich:
This is great, I've found myself needing nearly all these features.

Could you add a test for `largest_beam`?  I think we might also want a `smallest_beam`, maybe add that also?  I often want to know what my 'best' beam is because most of my slices will be at that 'best' resolution, but I'll have 1-2 outliers at the edges that will be 'largest' by a huge margin.

e-koch:
👍 

At some point, the multiple beam handling should probably be split into a `Beams` class in radio_beam. Especially for visualizing if/where outliers are occurring.

keflavich:
:+1: to a `Beams` class.  I often want to index `Beams` with a boolean array, which isn't possible for a list, and do other similar things (averaging, medianing, etc.)

Issue #364, opened 2017-02-17
Possible bug in spectral-cube - KeyError  in handle_beam_areas_wrapper


julian-garrido:
When opening a fits cube with spectral-cube and apply moment function I get the following error: 

```
from spectral_cube import SpectralCube as spc
....
cube = spc.read(image, fill_value=0, wcs="wcs")
cube_mask = cube.with_mask(cube > factor*rms * u.Jy)  
# compute 0th moment
mom0 = cube_mask.moment(order=0, how='cube')

Traceback (most recent call last):
  File "Moment0_LVM.py", line 56, in <module>
    mom0 = cube_mask.moment(order=0, how='cube')
  File "/Users/julian/anaconda/envs/LVM2/lib/python2.7/site-packages/spectral_cube/spectral_cube.py", line 2920, in newfunc
    or (hasattr(kwargs['axis'], '__len__') and
KeyError: 'axis'
```

could it be that the following code in spectral-cube should be modified as follows?
original code:
```
need_to_handle_beams = ('axis' in kwargs and kwargs['axis']==0
                                    or (hasattr(kwargs['axis'], '__len__') and
                                        0 in kwargs['axis'])
                                    or ('axis' not in kwargs and 'moment' in
function.__name__))
```
suggested code:
```
need_to_handle_beams = ('axis' in kwargs and 
                                    ( kwargs['axis']==0 or (hasattr(kwargs['axis'], '__len__') and
                                        0 in kwargs['axis']) )
                                    or ('axis' not in kwargs and 'moment' in
function.__name__))
```

keflavich:
@julian-garrido Would you like to submit this as a pull request?  I agree that the logic presented here needs a few extra parentheses to make the choices unambiguous.  It would be easier to review as a PR, and then your contribution will be recorded, but if you prefer not to, I can go ahead and make these changes.

We also need a test for this behavior.

keflavich:
Fixed by #366

PR #366, opened 2017-02-20
fix the logic in boolean expression to avoid KeyError - axis exception. #364

julian-garrido:


keflavich:
This is referencing Issue #364

PR #369, opened 2017-03-09
Flip the mask in spectral_interpolate

e-koch:


e-koch:
If `indiff < 0` when spectrally interpolating, the mask should also be flipped before interpolating.

I added a test for this case.

spectral_cube/spectral_cube.py
```diff
@@ -2519,8 +2519,10 @@ def spectral_interpolate(self, spectral_grid,
             cubedata = self.filled_data[::-1]
```
e-koch: @keflavich - Does the whole cube need to be loaded in like this? Is there an advantage to doing so?

keflavich: Hmm... no, now that you mention it, I think we could change this to
```
cubedata = self.filled_data
specslice = slice(None) if indiff >=0 else slice(None,None,-1)
```
then later, use
```
                newcube[:,iy,ix] = np.interp(spectral_grid.value, inaxis.value,
                                             cubedata[specslice,iy,ix].value)
```

keflavich:
Looks like there is one minor thing to fix or at least understand: Under some conditions, this warning isn't being raised?

https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/209818825#L704

e-koch:
I can reproduce it locally. Changing `warnings.simplefilter('default')` to `"always"` in `test_varyres_moment` and `test_varyres_moment_logic_issue364` seems to fix it. I'm unsure why this only happens for one of the test cases, though.

keflavich:
...is this affecting anything before your PR?

e-koch:
Yeah, it's failing on the master branch too. Looks like it has been for awhile: https://travis-ci.org/radio-astro-tools/spectral-cube/builds

keflavich:
We'll fix the other issue in #370

PR #376, opened 2017-04-05
HDU to Projection

e-koch:
Add creating a `Projection` from a FITS HDU.


keflavich:
I do this all the time.  Thanks!

keflavich:
Add a changelog entry?  Then we're good

e-koch:
👍

PR #377, opened 2017-04-21
add new docs for mask

keflavich:
this is to clarify the convention more clearly.

@DanRyanIrish does this help?


DanRyanIrish:
👍

PR #378, opened 2017-04-23
Add spatial functionality to projections

e-koch:
Adds much of the spatial functionality from `SpectralCube` for a 2D spatial `Projection`.

* Added `SpatialCoordMixinClass` for world coordinates handling.
* Beam handling/loading for `Projection`
* `Projection.convolve_to` and `Projection.reproject`
* `Projection.subimage` for slicing with quantities (as in `SpectralCube.subcube`)
* `WCSCelestialError` is raised if the `Projection` does not have 2 spatial axes
*  Added 2D test images

Not all of the spatial functionality has been abstracted out of `SpectralCube`. Things like `SpectralCube.flattened_world` require mask handling to be implemented for lower dimensional objects. Same goes for `SpectralCube.subcube_from_ds9_region`, in its current implementation.

### DIFF
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -3,6 +3,7 @@ py:obj radio_beam.Beam
 py:obj astroquery.splatalogue.Splatalogue
 py:class spectral_cube.base_class.BaseNDClass
 py:class spectral_cube.base_class.SpectralAxisMixinClass
+py:class spectral_cube.base_class.SpatialCoordMixinClass
 
 # yt references
 py:obj yt.surface.export_sketchfab
--- a/docs/nitpick-exceptions
+++ b/docs/nitpick-exceptions
@@ -9,3 +9,35 @@ py:class spectral_cube.base_class.SpatialCoordMixinClass
 py:obj yt.surface.export_sketchfab
 py:obj yt.show_colormaps
 
+# numpy inherited docstrings
+py:obj dtype
+py:obj a
+py:obj a.size == 1
+py:obj n
+py:obj ndarray
+py:obj args
+
+py:obj Quantity
+py:obj conjugate
+py:obj numpy.conjugate
+py:obj numpy.lib.stride_tricks.as_strided
+py:obj x
+py:obj i
+py:obj j
+py:obj axis1
+py:obj axis2
+py:obj numpy.ctypeslib
+py:obj ndarray_subclass
+py:obj ndarray.T
+py:obj order
+py:obj refcheck
+py:obj val
+py:obj offset
+py:obj lexsort
+py:obj a.transpose()
+py:obj new_order
+py:obj inplace
+py:obj subok
+py:obj ndarray.setflags
+py:obj ndarray.flat
+py:obj arr_t
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -56,8 +56,7 @@ class SpatialCoordMixinClass(object):
         of it.
 
         SpatialCoordMixinClass.world is called with *bracket notation*, like
-        a NumPy array::
-            c.world[0:3, :, :]
+        a NumPy array: ``c.world[0:3, :, :]``
 
         Returns
         -------
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -56,27 +56,33 @@ class SpatialCoordMixinClass(object):
         of it.
 
         SpatialCoordMixinClass.world is called with *bracket notation*, like
-        a NumPy array: ``c.world[0:3, :, :]``
+        a NumPy array::
+
+            c.world[0:3, :, :]
 
         Returns
         -------
         [v, y, x] : list of NumPy arrays
             The 3 world coordinates at each pixel in the view. For a 2D image,
-            the output is [y, x].
+            the output is ``[y, x]``.
 
 
         Examples
         --------
         Extract the first 3 velocity channels of the cube:
+
         >>> v, y, x = c.world[0:3]
 
-        Extract all the world coordinates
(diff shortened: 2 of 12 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -81,7 +80,7 @@ def wrapper(*args, **kwargs):
 # conventions between WCS and numpy
 np2wcs = {2: 0, 1: 1, 0: 2}
 
-class BaseSpectralCube(BaseNDClass, SpectralAxisMixinClass):
+class BaseSpectralCube(BaseNDClass, SpectralAxisMixinClass, SpatialCoordMixinClass):
```
keflavich: I just learned recently... I think `super` always refers to the *last* of these 3, and so maybe we want `BaseNDClass` last?  But we should check with someone who knows better (...or rtfd...) before changing it

spectral_cube/spectral_cube.py
```diff
@@ -183,20 +182,13 @@ def _new_cube_with(self, data=None, wcs=None, mask=None, meta=None,
...
 
-    def _try_load_beam(self, header):
+    def _attach_beam(self):
 
-        try:
-            from radio_beam import Beam
-        except ImportError:
-            warnings.warn("radio_beam is not installed. No beam "
-                          "can be created.")
+            beam = cube_utils.try_load_beam(self.header)
```
keflavich: is this now one level too far indented?

spectral_cube/base_class.py
```diff
@@ -31,6 +38,120 @@ def meta(self):
...
+        coordinates, strong distortions)
+
+        This method is not currently implemented. Use :meth:`world` instead.
+        """
+        raise NotImplementedError()
+
+    @property
+    def spatial_coordinate_map(self):
+
+        if self.ndim > 2:
```
keflavich: this is actually the case `if self.ndim == 3`.  Could do a general one:
```
view = [slice(0) for ii in range(self.ndim-2)] + [slice(None)]*2
return self.world[view][self.ndim-2:]
```

spectral_cube/base_class.py
```diff
@@ -31,6 +38,120 @@ def meta(self):
...
+        """
+        Returns a list of 1D arrays, for the world coordinates
+        along each pixel axis.
+
+        Raises error if this operation is ill-posed (e.g. rotated world
+        coordinates, strong distortions)
+
+        This method is not currently implemented. Use :meth:`world` instead.
+        """
+        raise NotImplementedError()
```
keflavich: This might be significantly more possible now if we can check `is_separable` on the wcs...

keflavich:
Also, I don't know what exactly went wrong with the docs, but something went very wrong.

spectral_cube/lower_dimensional_structures.py
```diff
@@ -246,6 +267,149 @@ def _quicklook_mpl(self, filename=None):
...
+
+        self._raise_wcs_no_celestial()
+
+        try:
+            from reproject.version import version
+        except ImportError:
+            raise ImportError("Requires the reproject package to be"
+                              " installed.")
+
+        # Need version > 0.2 to work with cubes
```
e-koch: I think so. We could just remove it here, or keep it so the requirements are homogeneous.

spectral_cube/base_class.py
```diff
@@ -1,16 +1,23 @@
...
 DOPPLER_CONVENTIONS = {}
 DOPPLER_CONVENTIONS['radio'] = u.doppler_radio
 DOPPLER_CONVENTIONS['optical'] = u.doppler_optical
 DOPPLER_CONVENTIONS['relativistic'] = u.doppler_relativistic
 
 
-
 class BaseNDClass(object):
+
+    _cache = {}
```
e-koch: Needed for caching. `BaseSpectralCube` does this as well: https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_cube.py#L133.

Switching to `lazyproperty` seems like the better route. Maybe that should be separate PR?

keflavich: Yes, separate PR is a good idea.

e-koch:
@keflavich -- The docs issue is from the LDOs inheriting from `Quantity`. For some reason, `doctest_skip` isn't ignoring the docstrings from `ndarray`. Any thoughts?

The docs *do* build fine, at least locally, with all of these warnings.

keflavich:
No immediate ideas, sorry.  Maybe someone in astropy will have a clue...

spectral_cube/lower_dimensional_structures.py
```diff
@@ -163,10 +165,11 @@ def quantity(self):
         """
         return u.Quantity(self)
 
-class Projection(LowerDimensionalObject):
+class Projection(LowerDimensionalObject, SpatialCoordMixinClass):
```
keflavich: What if you swapped the order of the inheritance?  Maybe that would fix the doc issues?

keflavich:
As @e-koch noted, this is related to https://github.com/astropy/astropy/pull/2159.  I wonder if there's a way to add numpy docs to the namespace, then the inherited docs wouldn't be a problem?  Otherwise, maybe we change something about docstring inheritance...?

docs/nitpick-exceptions
```diff
@@ -9,3 +9,35 @@ py:class spectral_cube.base_class.SpatialCoordMixinClass
 py:obj yt.surface.export_sketchfab
 py:obj yt.show_colormaps
 
+# numpy inherited docstrings
```
e-koch: 👍  for brute force

keflavich:
Everything passes now.  Good to merge @e-koch?

PR #380, opened 2017-04-27
Fill values for spectral_interpolate

low-sky:
`spectral_interpolate` defaults to the `np.interp` behaviour for extrapolates, which is to retain the first / last value in the domain.  Also, pardon the mess.  I'm trying out writing tests, so this might take a few tries.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2498,9 +2498,12 @@ class SpectralCube(BaseSpectralCube):
             If disabled, a warning will be raised when interpolating onto a
             grid that does not nyquist sample the existing grid.  Disable this
             if you have already appropriately smoothed the data.
-        fill_value : float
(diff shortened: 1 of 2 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -2498,10 +2498,14 @@ def spectral_interpolate(self, spectral_grid,
             If disabled, a warning will be raised when interpolating onto a
             grid that does not nyquist sample the existing grid.  Disable this
             if you have already appropriately smoothed the data.
+        fill_value : float
+            Value for cube interpolates outside of the original range
+            defined in the data
```
keflavich: I had a hard time parsing this; maybe `Value to use for extrapolation outside of the original image's domain.  The default is to use the nearest pixel in the original image.`  Is that correct, about the default?

low-sky:
I think we're good on the testing side.  Any other comments?

Issue #381, opened 2017-05-01
Cannot smooth spectrum generated from spectral_cube


jmangum:
I had hoped that the following would allow me to apply boxcar smoothing to a spectrum extracted using spectral cube:

```
cube2 = SpectralCube.read('IC860KbandAarrayHCNJ9v21.image.fits').with_spectral_unit(u.km/u.s,velocity_convention='optical',rest_value=restfreq2)

spectrum2 = cube2.subcube_from_ds9region(pyregion.open('IC860-20GHzbeam.reg')).mean(axis=(1,2))

spectrum2smooth = spectrum2.spectral_interpolate(cube1.spectral_axis)
```

Unfortunately, the last command yields the following error:

---------------------------------------------------------------------------
```
AttributeError                            Traceback (most recent call last)
<ipython-input-2-f8f473885b45> in <module>()
      3 spectrum2 = cube2.subcube_from_ds9region(pyregion.open('IC860-20GHzbeam.reg')).mean(axis=(1,2))
      4
----> 5 spectrum2smooth = spectrum2.spectral_interpolate(cube1.spectral_axis)

/Users/jmangum/anaconda/lib/python2.7/site-packages/astropy/units/quantity.pyc in __getattr__(self, attr)
    861                 "'{0}' object has no '{1}' member".format(
    862                     self.__class__.__name__,
--> 863                     attr))
    864
    865         def get_virtual_unit_attribute():

AttributeError: 'OneDSpectrum' object has no 'spectral_interpolate' member
```

e-koch:
Added in #382

PR #382, opened 2017-05-02
add spectral_interpolate to OneDSpectrum and fix the with_spectral_unit component a little

keflavich:
Addresses #381

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -356,8 +356,10 @@ class OneDSpectrum(LowerDimensionalObject,SpectralAxisMixinClass):
                                                  rest_value=rest_value)
 
         newheader = self._nowcs_header.copy()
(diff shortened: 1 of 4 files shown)


spectral_cube/lower_dimensional_structures.py
```diff
@@ -387,3 +389,100 @@ def hdulist(self):
...
+
+    def spectral_interpolate(self, spectral_grid,
+                             suppress_smooth_warning=False):
+        """
+        Resample the spectrum onto a specific grid
+
+        Parameters
+        ----------
+        spectral_grid : array
+            An array of the spectral positions to regrid onto
```
astrofrog: I wouldn't call it regridding, as that means something different to me than interpolation

keflavich: You're right; we should probably call the function `spectral_regrid`, since that's closer to what it's doing.

PR #386, opened 2017-05-16
Rename 'interpolate' functions to 'regrid'

keflavich:
The 'interpolate' functions are all doing regridding, so that's a better name for them.

cc @astrofrog

### DIFF
--- a/docs/smoothing.rst
+++ b/docs/smoothing.rst
@@ -49,9 +49,9 @@ Example::
     kernel = Gaussian1DKernel(2.5)
     new_cube = cube.spectral_smooth(kernel)
 
-This can be useful if you want to interpolate onto a coarser grid but maintain
+This can be useful if you want to regrid onto a coarser grid but maintain
 Nyquist sampling.  You can then use the
-`~spectral_cube.SpectralCube.spectral_interpolate` method to regrid your
+`~spectral_cube.SpectralCube.spectral_regrid` method to regrid your
 smoothed spectrum onto a new grid.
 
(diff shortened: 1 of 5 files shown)


low-sky:
Hmmm.  I find _interpolate_ to be a more precise name.  _Regridding_ comes with connotations of behaving well statistically, whereas you know you are screwing something up by interpolating.  Current implementations in spectral_cube don't behave well.  I have a student working on a Fourier implementation.

That being said, the right move here may be to leave `regrid` as per this PR and then stick in a `method='interpolate'` as the default and then use `method='fourier'` for later.

keflavich:
I like the approach.  I think @astrorog's point was that the interpolation being done is regridding, in the sense that the output is not on the same grid as the input.  The same is true of interpolation, though...

:+1: to Fourier upsampling, though!  That's my favorite!

see https://github.com/keflavich/image_registration/blob/master/image_registration/fft_tools/upsample.py

keflavich:
on fourier upsampling, this approach is probably nicer:
https://github.com/keflavich/image_registration/blob/master/image_registration/fft_tools/zoom.py#L6
https://github.com/keflavich/image_registration/blob/master/image_registration/fft_tools/scale.py#L4

(it looks like I may have done the math for those... so best to double check it)

keflavich:
@astrofrog, @low-sky because `interpolate` and `regrid` each seem to carry different implicational baggage depending on who you ask, I'm going to leave this as `interpolate` for now - technically, that is what is being done under the hood (`np.interp`), and interpolation without regridding is probably nonsense?

PR #394, opened 2017-06-01
More fixes to the LDO mask handling, etc.

keflavich:
This revisits issue #392, which wasn't actually solved by that PR.  However, it
turns out we need to make deeper changes.

Issue #367, opened 2017-02-28
moment() drops the 3rd axis [or: cubes must have a 3rd axis > 1]


teuben:
Since moment() always drops information on the 3rd axis, I cannot read the results back.  Is this an intended design, or meant to be resolvable?

keflavich:
What information is missing?  You mean the frequency axis information?  I think you're right, that should be preserved in the header.  Can it reasonably be put into CRPIX/CRVAL/CDELT, or will most programs misinterpret that as indicating the presence of a cube?

teuben:
yes, NAXIS=2.   I'd argue we could try NAXIS=3 and NAXIS3=1 and use an appropriate description for the compressed CRVAL/CDELT.  Both MIRIAD and CASA do it that way.  Sounds like a DtU17 hackday option.

low-sky:
I think there's a larger question here.  Moments are 2D images and thus should they be read into spectral-cube?  By default, I would say "this is working as intended.  go use astropy.fits for dealing with images"  However, moment0 straddle that case where you could talk about a 1 plane image with bandwidth equal to the original image.  Moment1++ isn't a good fit to this 3D cubes though.

keflavich:
I've never been a fan of `NAXIS=3 NAXIS3=1` because it is a somewhat inaccurate representation of the data; for example, optical images with wide- or narrow-band filters usually do not have any "fake" third dimension.  However, I understand why you want it.

Maybe we can provide an option for writing out in "cube" or "flat" format?  I would default to "flat".  We'd implement this in a special-case `Moment0Projection` `LowerDimensionalObject`.

teuben:
ok, that's what i needed to hear :-) moments are 2D images, not cubes (but as I said, both casa and miriad can do a little more with them since they do retain some useful compressed information).

low-sky:
@teuben What sort of functionality, apart from the obvious read/write, would you be looking for in `SpectralCube` operating on saved moments?

e-koch:
We already carry some meta-data along in the moments. Perhaps it can be kept there? Along with the spectral axis information, things like the spectral range used to create the moment 0 would be useful too.

low-sky:
[...]

Issue #111, opened 2014-06-20
Invalid WCS when saving moment0 maps of non-spectral axes


ChrisBeaumont:
Works:

```
cube.moment0(axis=0).write('test.fits')
cube.moment0(axis=1)
```

Fails:

```
In [13]: s.moment0(axis=1).write('test.fits')
ERROR: InconsistentAxisTypesError: ERROR 4 in wcs_types() at line 1973 of file cextern/wcslib/C/wcs.c:
Unmatched celestial axes.
 [astropy.wcs.wcs]
---------------------------------------------------------------------------
InconsistentAxisTypesError                Traceback (most recent call last)
<ipython-input-13-2419fce08bed> in <module>()
----> 1 s.moment0(axis=1).write('test.fits')

/Users/beaumont/alma/spectral-cube/spectral_cube/spectral_cube.pyc in write(self, filename, format, overwrite)
    124             format = determine_format(filename)
    125         if format == 'fits':
--> 126             self.hdu.writeto(filename, clobber=overwrite)
    127         else:
    128             raise ValueError("Unknown format '{0}' - the only available "

/Users/beaumont/alma/spectral-cube/spectral_cube/spectral_cube.pyc in hdu(self)
    104             hdu = fits.PrimaryHDU(self.value)
    105         else:
--> 106             hdu = fits.PrimaryHDU(self.value, header=self.wcs.to_header())
    107         hdu.header['BUNIT'] = self.unit.to_string(format='fits')
    108         return hdu

/Users/beaumont/anaconda/lib/python2.7/site-packages/astropy/wcs/wcs.pyc in to_header(self, relax, key)
   1663 
   1664         if self.wcs is not None:
-> 1665             header_string = self.wcs.to_header(relax)
   1666             header = fits.Header.fromstring(header_string)
   1667         else:

InconsistentAxisTypesError: ERROR 4 in wcs_types() at line 1973 of file cextern/wcslib/C/wcs.c:
Unmatched celestial axes.
```

keflavich:
The more we dig into slicing, projecting, etc., the more it appears we need to preserve the full WCS when slicing.  A 1D slice should have a length-1 coordinate for each of the sliced-out coordinates, and so on.  This is fine and pretty easy for slices, but nontrivial for projections (what coordinate do you use when taking the `max` along an axis?).

[...]

PR #395, opened 2017-06-03
WIP: Keep full 3D WCS objects for LDOs

keflavich:
Related to #111, #367, #392, #394:
When slicing or projecting a cube, there is generally still a 3rd dimension,
even if it is not reflected in the data.  Currently, if you attempt to do
anything with the WCS of a position-velocity slice, the WCS crashes, and we
want to avoid that.

There are big open questions though:

 - for projections, what coordinate applies to operations like `.max`?
 - for projections, can we assume the central pixel (the 'mean' along that
 axis) is the appropriate coordinate?
 - What should CDELT be?

in slices, I think all of these questions are easily answered.

We will need to provide convenience functions (WCS wrappers) to get the
coordinates in projections and slices, because most users will never care about
the sliced-over axes, or they will care but only want to see them as metadata.

cc @e-koch, who has contributed significantly to this discussion already.


keflavich:
Tests are pretty close to passable, but they fail for the aplpy case because now we're writing out HDUs with 3 dimensions for a 2 dimensional object.  We just need to tell that to aplpy in the vis calls...

keflavich:
We really need this, unfortunately.

Here's an example case I had to do to get WCSes attached properly:
```python
for fn in cubefns:
    cube = SpectralCube.read(fn).subcube_from_regions([reg])
    print(cube)
    avg = cube.mean(axis=(1,2))
    hdu = avg.hdu
    y,x = cube.wcs.celestial.world_to_pixel(reg.center)

    slc = getslice()[int(floor(x)):int(ceil(x)), int(floor(y)):int(ceil(y)), :]
    ww = wcs_utils.slice_wcs(cube.wcs, slc, numpy_order=False)
    hdu.header.update(ww.to_header())
    hdu.data = hdu.data[None,None,:]
```

PR #396, opened 2017-06-06
Another fix to 1D masking

keflavich:
When you index a masked object with a boolean array (or a single value), you should get NaN if the value was masked out.


keflavich:
cc @e-koch any thoughts?  I'll merge quickly otherwise.

e-koch:
:+1:

PR #400, opened 2017-06-28
Fix OneDSpectrum masking through interpolation

e-koch:
Added mask handling for `OneDSpectrum.spectral_interpolate`. The mask has to be handled for `spectral_interpolate` to work since we now check whether the WCS in the mask is valid with the data.

Also:
* `MaskBase.quicklook` was converting to a Projection to view, but this is won't work when we need to import mask classes into the LDOs file. The quicklook is now a near copy-paste of the quicklook in `Projection`. I could add a general function to do this, but it's not a particularly important few lines... 
* `OneDSpectrum` checks the mask type when a new spectrum is made.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,14 @@ env:
         - ASTROPY_VERSION=stable
         - CONDA_DEPENDENCIES='matplotlib aplpy'
         - CONDA_CHANNELS='astropy-ci-extras astropy'
-        - PIP_DEPENDENCIES='pytest-xdist pvextractor'
+        - PIP_DEPENDENCIES='pvextractor'
         - SETUP_XVFB=True
 
     matrix:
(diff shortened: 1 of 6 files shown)


e-koch:
Weird error on some of the travis jobs: https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/248024982

keflavich:
that was weird.  restarted.

e-koch:
I tried restarting a couple and they failed again.

keflavich:
I don't understand these failures.  They look like a problem in the conda-based build environment.  Maybe we can lock pytest to a specific version?  This sort of failure will probably start showing up on other astropy-helpers-based packages

e-koch:
pytest is the same version. I think pytest-xdist is the issue. Do we even need it installed? We're not running each test case in parallel.

PR #402, opened 2017-07-03
Try to fix pytest-xdist issue

keflavich:
cc @e-koch - I was able to reproduce a failure by pip installing pytest-xdist locally.  It wasn't the same failure, but.... close enough?  pip installing the latest pytest seemed to fix it


e-koch:
:+1:

keflavich:
Hooray!!

PR #405, opened 2017-07-10
Add a document to help users figure out how to deal with commonly-encountered errors

keflavich:
Documentation needs a lot of expansion.  This is a step in the right direction.

### DIFF
--- a/docs/errors.rst
+++ b/docs/errors.rst
@@ -1,3 +1,6 @@
+.. doctest-skip-all
+
+
 Explanations of commonly-encountered error messages
 ===================================================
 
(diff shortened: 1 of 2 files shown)


keflavich:
Rendered version:
http://keflavich-spectralcube.readthedocs.io/en/error_docs/errors.html

cc @e-koch @low-sky if you come up with any more of these, please add.   Things people have mentioned recently:

 * [ ] Sometimes moment2 results in NaNs when moment1 and moment0 both work.  This is usually because of bad baseline subtraction.  It has affected at least 3 individuals so far

(let's add to this checklist)

PR #410, opened 2017-09-04
Add tools to load a OneDSpectrum from a FITS HDU

keflavich:
TODO: add a test


keflavich:
There are very few new lines that aren't tested, and they're corner cases anyway.

Issue #415, opened 2017-09-08
Add radio_beam as a dependency


astrofrog:
I would like to suggest that we add radio_beam in ``setup.py`` now that there is a release. I don't think it will stop non-radio users from using the package, it's simply that we use this dependency to support radio cubes. And radio-beam is a pretty lightweight dependency!

keflavich:
I'm OK with this.  A PR would be welcome.  I'll try to issue a new release of spectral-cube next week...

olebole:
Hmm, this means that I need to package radio_beam now before I can update spectral_cube in Debian.

PR #421, opened 2017-09-13
Added radio_beam as a required dependency

astrofrog:
Fixes https://github.com/radio-astro-tools/spectral-cube/issues/415

I think this is acceptable because it is such a lightweight dependency and the warnings are so annoying when reading in spectral cubes and not having radio_beam installed.


keflavich:
It looks like specifying `install_requires=[..., 'radio_beam']` wasn't enough to get it installed?  Is this a .travis bug or a real bug in the setup.py file?

astrofrog:
Huh it did install but is not importable: https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/275127368 - will investigate

Issue #420, opened 2017-09-13
Give more helpful error if WCS is not spatial + spectral


astrofrog:
I read in a cube that had invalid WCS, and the error was:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-80fad4281d1f> in <module>()
----> 1 cube = SpectralCube.read('JWST-NIRSpec-IFU-GAL2136_CLEAR_PRISM.fits', hdu=1)

~/miniconda3/envs/dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in read(cls, filename, format, hdu, **kwargs)
   2026         from .io.core import read
   2027         from .stokes_spectral_cube import StokesSpectralCube
-> 2028         cube = read(filename, format=format, hdu=hdu, **kwargs)
   2029         if isinstance(cube, StokesSpectralCube):
   2030             if hasattr(cube, 'I'):

~/miniconda3/envs/dev/lib/python3.6/site-packages/spectral_cube/io/core.py in read(filename, format, hdu, **kwargs)
     31     if format == 'fits':
     32         from .fits import load_fits_cube
---> 33         return load_fits_cube(filename, hdu=hdu, **kwargs)
     34     elif format == 'casa_image':
     35         from .casa_image import load_casa_image

~/miniconda3/envs/dev/lib/python3.6/site-packages/spectral_cube/io/fits.py in load_fits_cube(input, hdu, meta, **kwargs)
    143     if wcs.wcs.naxis == 3:
    144 
--> 145         data, wcs = cube_utils._orient(data, wcs)
    146 
    147         mask = LazyMask(np.isfinite, data=data, wcs=wcs)

~/miniconda3/envs/dev/lib/python3.6/site-packages/spectral_cube/cube_utils.py in _orient(array, wcs)
    151         raise ValueError("Input WCS should not contain stokes")
    152 
--> 153     t = [types.index('spectral'), nums.index(1), nums.index(0)]
    154     result_array = array.transpose(t)
    155 

ValueError: 'spectral' is not in list
```

I think we should have a clearer error message that says 'The WCS in HDU=... does not include valid spatial or spectral coordinates'

PR #425, opened 2017-09-14
Improve exception if cube does not have two celestial and one spectral axis

astrofrog:
Fixes https://github.com/radio-astro-tools/spectral-cube/issues/420


keflavich:
Looks fine, so I'll merge, but it wouldn't hurt to have a test

PR #429, opened 2017-09-22
Added in spatial and median filtering

brechmos-stsci:
I added in spatial filtering which is very similar to the spectral filtering function. This uses the same mechanism of 2D kernels defined in astropy.
I also added in 2D spatial median filtering. 

I ran some basic tests on a dataset I have:
image

CC: @astrofrog

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2423,10 +2423,7 @@ class SpectralCube(BaseSpectralCube):
             pb.update()
 
             if includemask.any():
-                if str(kernel) == 'median':
-                    return ndimage.filters.median_filter(im, size=3)
-                else:
-                    return convolve(im, kernel, normalize_kernel=True, **kwargs)
+                return convolve(im, kernel, normalize_kernel=True, **kwargs)
             else:
                 return im
 
@@ -2442,6 +2439,70 @@ class SpectralCube(BaseSpectralCube):
 
         # TODO: do something about the mask?
         newcube = self._new_cube_with(data=smoothcube_, wcs=self.wcs,
+                                      mask=self.mask, meta=self.meta,
+                                      fill_value=self.fill_value)
+
+        return newcube
+
+    def spectral_smooth_median(self, ksize,
+                        #numcores=None,
+                        **kwargs):
+        """
+        Smooth the cube along the spectral dimension
+
+        Parameters
+        ----------
+        kernel : `~astropy.convolution.Kernel1D`
+            A 1D kernel from astropy
+        convolve : function
+            The astropy convolution function to use, either
+            `astropy.convolution.convolve` or
+            `astropy.convolution.convolve_fft`
+        kwargs : dict
+            Passed to the convolve function
+        """
+
+        shape = self.shape
+
+        # "cubelist" is a generator
+        # the boolean check will skip smoothing for bad spectra
+        # TODO: should spatial good/bad be cached?
+        cubelist = ((self.filled_data[:,jj,ii],
+                     self.mask.include(view=(slice(None), jj, ii)))
+                    for jj in range(self.shape[1])
+                    for ii in range(self.shape[2]))
+
+        pb = ProgressBar(shape[1]*shape[2])
+
+        def _gsmooth_spectrum(args):
+            """
+            Helper function to smooth a spectrum
+            """
+            (spec, includemask),kwargs = args
+            pb.update()
+
+            if any(includemask):
+                return ndimage.filters.median_filter(spec, size=ksize)
+            else:
+                return spec
+
+        # could be numcores, except _gsmooth_spectrum is unpicklable
+        with cube_utils._map_context(1) as map:
+            smoothcube_ = np.array([x for x in
+                                    map(_gsmooth_spectrum, zip(cubelist,
+                                                               itertools.cycle([kwargs]),
+                                                              )
+                                       )
+                                   ]
+                                  )
+
+        # empirical: need to swapaxes to get shape right
+        # cube = np.arange(6*5*4).reshape([4,5,6]).swapaxes(0,2)
+        # cubelist.T.reshape(cube.shape) == cube
+        smoothcube = smoothcube_.T.reshape(shape)
+
+        # TODO: do something about the mask?
+        newcube = self._new_cube_with(data=smoothcube, wcs=self.wcs,
                                       mask=self.mask, meta=self.meta,
                                       fill_value=self.fill_value)
 
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2341,6 +2341,8 @@ class SpectralCube(BaseSpectralCube):
         kwargs : dict
             Passed to the convolve function
         """
+        if not scipyOK:
+            raise ImportError("Scipy could not be imported: this function won't work.")
 
         shape = self.shape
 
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
I added a second commit that added a function to run a median filter in the spectral direction.
image

brechmos-stsci:
If there is a wish for me to rework the interface for the median filtering, I am happy to do it.

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+    def spatial_smooth_median(self, ksize,
+                        #numcores=None,
+                        **kwargs):
+        """
+        Smooth the cube along the spectral dimension
+
+        Parameters
+        ----------
+        kernel : `~astropy.convolution.Kernel1D`
+            A 1D kernel from astropy
```
keflavich: Spatial smooth should use a 2D kernel, no?

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+        kwargs : dict
+            Passed to the convolve function
+        """
+
+        shape = self.shape
+
+        # "cubelist" is a generator
+        # the boolean check will skip smoothing for bad spectra
+        # TODO: should spatial good/bad be cached?
+        cubelist = ((self.filled_data[ii],
```
keflavich: this is a list of 2D images, no?  so `imagelist` or `planelist` would be better?

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+                     self.mask.include(view=(ii, slice(None), slice(None))))
+                    for ii in range(self.shape[0]))
+
+        pb = ProgressBar(shape[0])
+
+        def _gsmooth_image(args):
+            """
+            Helper function to smooth a spectrum
```
keflavich: an image?

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+            if includemask.any():
+                return ndimage.filters.median_filter(im, size=ksize)
+            else:
+                return im
+
+        # could be numcores, except _gsmooth_spectrum is unpicklable
+        with cube_utils._map_context(1) as map:
+            smoothcube_ = np.array([x for x in
+                                    map(_gsmooth_image, zip(cubelist,
+                                                               itertools.cycle([kwargs]),
```
keflavich: whitespace here; the indentation of `itertools` should match that of `cubelist`

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+                        #numcores=None,
+                        convolve=convolution.convolve,
+                        **kwargs):
+        """
+        Smooth the cube along the spectral dimension
+
+        Parameters
+        ----------
+        kernel : `~astropy.convolution.Kernel1D`
+            A 1D kernel from astropy
```
keflavich: ditto above: 2D

spectral_cube/spectral_cube.py
keflavich: In the median case, there is no kernel, so this should be removed.

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+    def spectral_smooth_median(self, ksize,
+                        #numcores=None,
+                        **kwargs):
+        """
+        Smooth the cube along the spectral dimension
+
+        Parameters
+        ----------
+        kernel : `~astropy.convolution.Kernel1D`
+            A 1D kernel from astropy
```
keflavich: no kernel....

spectral_cube/spectral_cube.py
```diff
@@ -2328,6 +2328,186 @@ def _new_cube_with(self, **kwargs):
...
+        convolve : function
+            The astropy convolution function to use, either
+            `astropy.convolution.convolve` or
+            `astropy.convolution.convolve_fft`
```
keflavich: no convolve...

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

spectral_cube/spectral_cube.py
brechmos-stsci: fixed and push coming shortly

brechmos-stsci:
@keflavich I see some fits files referenced in the tests (e.g., adv.fits, vda_beams.fits etc).  Do you have a 3D cube that I should use for the tests, and if so, how could I access it to test if it is not in the `tests/data` directory?

keflavich:
@brechmos-stsci If you run `python setup.py install` (or `python setup.py [anything]`...), it should create cubes in the `tests/data` directory.  If that fails for some reason, raise an Issue, but you can also run `make` in the `tests/data` directory and it will build the test cubes.

brechmos-stsci:
@keflavich I made changes to the code and added several tests to one of the test files.  The tests all appear to hit the `ImportError` due to scipy.ndimage not being available.  Is there something I need to do in the tests in order to flag that it is needed?
```
        if not scipyOK:
>           raise ImportError("Scipy could not be imported: this function won't work.")
E           ImportError: Scipy could not be imported: this function won't work.
```
I am not overly experienced with Travis yet. :-/

keflavich:
Yeah, in the tests you can do:

```python
import pytest
try:
    import scipy.ndimage
    SCIPYOK = True
except ImportError:
    SCIPYOK = False

pytest.skipif('not SCIPYOK')
def test_that_relies_on_scipy():
     pass
```

brechmos-stsci:
@keflavich The two tests that failed, now, are due to import errors trying to do `import signature` which is not related to code I worked on (unless there is some connection I am not aware of).  So, I believe this should be good, but please peruse again. :-)

keflavich:
Oh, I think this issue is that we need to remove py2.7+astropy-dev from the travis build matrix.

keflavich:
@brechmos-stsci I just merged in a test grid update in #431.  Could you update against that (either merge or rebase)?  That should fix the tests

brechmos-stsci:
I merged your test grid update changes and the Travis tests ran. If there are any other changes / updates you would like, please tag me. Happy to make them.

Issue #432, opened 2017-10-03
Convolution of a data cube produced by a radiative transfer code


smaret:
I'm trying to use the `.convolve_to()` method to convolve a cube produced by a radiative transfer code, but I'm getting the following error:

```python
>>> cube = SpectralCube.read("CO.fits.gz")
WARNING: FITSFixedWarning: 'spcfix' made the change 'Changed CTYPE3 from 'VELO-LSR' to 'VOPT', and SPECSYS to 'LSRK' (VELREF=0)'. [astropy.wcs.wcs]
/usr/local/lib/python2.7/site-packages/spectral_cube/spectral_cube.py:246: UserWarning: Could not parse beam information from header.  Exception was: TypeError('No BMAJ found and does not appear to be a CASA/AIPS header.',)
  "  Exception was: {0}".format(ex.__repr__()))
>>> cube = cube.convolve_to(beam)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/spectral_cube/spectral_cube.py", line 2630, in convolve_to
    convolution_kernel = beam.deconvolve(self.beam).as_kernel(pixscale)
AttributeError: 'SpectralCube' object has no attribute 'beam'
```

Indeed, the data cube has no beam information since it comes from a radiative transfer code. But why a cube needs to have a beam attribute to be convolved in the first place?

keflavich:
The `convolve_to` function convolves to a specified resolution; in order to
do so, it must know the original resolution.  We can expose another
function `convolve_with` which would convolve with any user- specified
kernel.  I'll implement that.

e-koch:
@keflavich - `SpectralCube.spatial_smooth` should do that already. The docs don't mention it yet, though.

But it would be handy to be able to pass your own `Beam` into `SpectralCube` too. Starting with a point-like beam seems like a common starting point with synthetic observations.

keflavich:
Oh, right, thanks.

And yes, a pointlike beam would be nice to have available... I don't know if that works with the deconvolve tools.

keflavich:
Apparently the docstring doesn't even show up in our docs, but this is the relevant function: https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_cube.py#L2384

smaret:
Thanks for your answers. I agree it would be nice to be able to pass a 'Beam' instance to a `SpectralCube`.

[...]

PR #433, opened 2017-10-03
Attach custom beam

e-koch:
From #432. A beam object can now be attached with:
```
newcube = cube.with_beam(beam)
newproj = projection.with_beam(beam)
```

Since radio-beam is now a dependency, I've removed the `read_beam` option from `SpectralCube`. We need to keep it for `Projection` though since there remains the issue with non-matching celestial axes for some slices.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2291,7 +2291,6 @@ class SpectralCube(BaseSpectralCube):
             self._meta['beam'] = beam
             self._header.update(beam.to_header_keywords())
 
-        if 'beam' in self._meta:
             self.pixels_per_beam = (self.beam.sr /
                                     (astropy.wcs.utils.proj_plane_pixel_area(self.wcs) *
                                      u.deg**2)).to(u.dimensionless_unscaled).value
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2679,8 +2679,7 @@ class SpectralCube(BaseSpectralCube):
                                        normalize_kernel=True)
             pb.update()
 
-        newcube = self._new_cube_with(data=newdata, read_beam=False,
-                                      beam=beam)
+        newcube = self._new_cube_with(data=newdata, beam=beam)
 
         return newcube
 
@@ -3213,7 +3212,6 @@ class VaryingResolutionSpectralCube(BaseSpectralCube):
                                meta=self.meta, fill_value=self.fill_value,
                                header=self.header,
                                allow_huge_operations=self.allow_huge_operations,
-                               read_beam=False,
                                beam=beam,
                                wcs_tolerance=self._wcs_tolerance)
 
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2853,6 +2853,16 @@ class VaryingResolutionSpectralCube(BaseSpectralCube):
 
             # Slice objects know how to parse Beam objects stored in the
             # metadata
+            # A 2D slice with a VRSC should not be allowed along a
+            # position-spectral axis
+            if not isinstance(self.beams[specslice], Beam):
+                raise AttributeError("2D slices along a spectral axis are not "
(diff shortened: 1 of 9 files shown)


keflavich:
I would prefer this be a `with_beam` method; nearly everything else we do related to modifying cubes forces the creation of a new cube object, and it feels weird having a one-off `attach` method.  If a beam is really needed as part of the original cube, it should be established during instantiation.

spectral_cube/spectral_cube.py
```diff
@@ -2292,7 +2279,17 @@ def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
...
         # Beam loading must happen *after* WCS is read
-        self.attach_beam(beam=beam)
+
+        if beam is None:
+            beam = cube_utils.try_load_beam(self.header)
+        else:
+            if not isinstance(beam, Beam):
+                raise TypeError("beam must be a radio_beam.Beam object.")
+
+        if beam is not None:
```
keflavich: we're already in the `else` of `if beam is None`, so this `if` could just be deleted with no change of indentation, I think

e-koch:
Agreed. I changed both to be `with_beam` and return a new object.

It highlighted one issue in 2D objects when the slice is taken along a position-velocity slice:
```
___________________________ TestSpectralCube.test_getitem[vda_beams-trans8] ____________________________

self = <spectral_cube.tests.test_spectral_cube.TestSpectralCube object at 0x7f33e9101190>
name = 'vda_beams', trans = [0, 2, 1]

    @pytest.mark.parametrize(('name', 'trans'), translist)
    def test_getitem(self, name, trans):
        c, d = cube_and_raw(name + '.fits')
    
        expected = np.squeeze(d.transpose(trans))
    
        assert_allclose(c[0,:,:].value, expected[0,:,:])
>       assert_allclose(c[:,:,0].value, expected[:,:,0])

spectral_cube/tests/test_spectral_cube.py:257: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
spectral_cube/spectral_cube.py:2862: in __getitem__
    meta=meta)
spectral_cube/lower_dimensional_structures.py:236: in __new__
    self.beam = beam
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Slice [[ 0.54671028, 0.96958463, 0.93949894],
        [ 0.59789998, 0.0884925...2873751, 0.28093451],
        [ 0.14092422, 0.07455064, 0.77224477]] Jy / beam>
obj = [Beam: BMAJ=0.10000000149 arcsec BMIN=0.40000000596 arcsec BPA=0.0 deg, Beam: BMAJ=0.20000000298 arcsec BMIN=0.3000000... arcsec BMIN=0.20000000298 arcsec BPA=60.0 deg, Beam: BMAJ=0.40000000596 arcsec BMIN=0.10000000149 arcsec BPA=30.0 deg]

    @beam.setter
    def beam(self, obj):
    
        if not isinstance(obj, Beam):
>           raise TypeError("beam must be a radio_beam.Beam object.")
E           TypeError: beam must be a radio_beam.Beam object.
```
We should probably just not allow this type of slicing with a VRSC.

spectral_cube/spectral_cube.py
e-koch: no, I'll have to duplicate the three lines below that `if` without it for when the beam is loaded from the header. The other option would be to move loading a beam from the header somewhere into `io.fits`

keflavich: right, good point, OK.

e-koch:
@keflavich - separate PR for the above error? One test should fail on travis.

keflavich:
It's OK to fix it here if you want

PR #441, opened 2017-10-30
Channel numbering in multibeam cubes should be sequential

keflavich:
The "channel" column of the beam table seems to be relatively useless, as it is just an index that is required to match `range(len(table))`.


keflavich:
Confirmed that this fixes an issue noted by Ciriaco.

PR #442, opened 2017-10-30
Bugfix: xhi < xlo results in bad WCS

keflavich:
This fixes a bug in which `subcube` could create a bad WCS by slicing from high
to low instead of low to high.  This issue can arise when slicing in RA, which
usually has CDELT < 0, such that a lower pixel number maps to a higher value in
degrees.


keflavich:
cc @nithyanandan - thanks for finding this

keflavich:
the first commit above was already merged in #441

PR #446, opened 2017-11-20
Add stacking tools

keflavich:
I've added a spectral stacking tool - stacking in spatial dimensions to create a stacked spectrum - and will soon add a spectral stacking tool also.  

Question - where should these live?  @adrn, I've put them in analysis_utilities (...not to be abbreviated), which I think is a descriptive location, but what name would you recommend here?

@e-koch, @low-sky - you might have useful suggestions on naming / testing / other things to add to this?

### DIFF
--- a/spectral_cube/analysis_utilities.py
+++ b/spectral_cube/analysis_utilities.py
@@ -1,0 +1,50 @@
+import numpy as np
+
+from astropy import units as u
+
+def stack_spectra(cube, velocity_field, average=np.nanmean):
+    """
+    Stack all spectra in a cube by shifting each spectrum in velocity and then
+    averaging
+
+    Parameters
+    ----------
+    cube : SpectralCube
+        The cube
+    velocity_field : Quantity
+        A Quantity array with m/s or equivalent units
+    average : function
+        A function that can operate over a list of numpy arrays (and accepts
+        ``axis=0``) to average the spectra.  `numpy.nanmean` is the default,
+        though one might consider `numpy.mean` or `numpy.median` as other
+        options.
+
+    """
+
+    if not all(cube.shape[1:] == velocity_field.shape):
+        raise ValueError("Cube spatial dimensions must match velocity field map dimensions")
+
+    if not cube.spectral_axis.unit.is_equivalent(u.m/u.s):
+        # select the mean velocity/wavelength as the reference
+        ref = cube.spectral_axis.mean()
+        vcube = cube.with_spectral_unit(u.km/u.s, velocity_convention='radio', rest_value=ref)
+    else:
+        vcube = cube
+
+    velax = vcube.spectral_axis
+
+    spectra = []
+
+    for index, velocity in np.ndenumerate(velocity_field):
+        if np.isfinite(velocity):
+            shifted_vel = velax - velocity
+            newspec = np.interpolate(shifted_vel, velax, vcube[:, index[0], index[1]])
+            spectra.append(newspec)
+
+    meanspec = average(spectra, axis=0)
+
+    template_hdu = cube.with_spectral_unit(u.GHz)[:,0,0].hdu
+
+    template_hdu.data = meanspec
+
+    return template_hdu
--- a/spectral_cube/analysis_utilities.py
+++ b/spectral_cube/analysis_utilities.py
@@ -3,8 +3,8 @@ import numpy as np
 from astropy import units as u
 from astropy.extern.six.moves import zip
 from astropy.extern.six.moves import range as xrange
-from astropy import log
 from astropy.wcs import WCS
+from astropy.utils.console import ProgressBar
 
 from .cube_utils import _map_context
 from .lower_dimensional_structures import OneDSpectrum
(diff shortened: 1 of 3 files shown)


e-koch:
I have an implementation for the shifting hidden away [here](https://github.com/e-koch/CubeAnalysis/blob/master/cube_analysis/spectra_shifter.py). That repo has become the place for tools that I didn't seem to fit into existing packages (and a lot is experimental). You're welcome to take whichever of that code will work here (and I'm happy to help, too).

There's a couple of [basic tests](https://github.com/e-koch/CubeAnalysis/blob/master/cube_analysis/tests/test_shifter.py) that can be adapted to work here, too.

I prefer doing the shifting in the Fourier plane since it minimizes the effect of finite channel sizes and the need to interpolate onto finer grids.

keflavich:
Thanks.  Fourier shifting is nice, but it's probably no better or worse than direct interpolation when you're working within the same cube.

Your code has a place in spectral-cube.  Want to try to merge it in with this?  Possibly replace what I did and bring in your tests & all?

low-sky:
I strongly advocate Fourier shifting since it preserves the noise properties of the spectrum for sub-channel shifts.

e-koch:
Sure. I'll clean/refactor somewhat before adding to the PR.

keflavich:
Presumably it preserves the noise properties because the spectrum was already created with an FT?  Is there any reason to believe direct interpolation does not?
(I have no problem using the FT-based shift, just trying to understand)

low-sky:
Direct interpolation can be done right, but it's essentially Fourier interpolation (i.e., you have to interpolate over all the points anyways).  If you do something like a naive linear interpolation of half a channel shift, you put your output data as the average of the two points on either side: x_new = (x_i + x_(i+1))/2, which basically smooths and your noise drops by sqrt(2).  For free.  As noted, you can do better, but it seems like the complexity gains approach FTs.

e-koch:
@keflavich - I added a refactored version of my stacking functions and some simple tests.

I didn't touch `stack_cube`. I'm not sure if you meant for it to average between multiple lines, or shuffle and then average them.

e-koch:
A bunch of the doctests are failing. Seems to be unrelated to this PR

spectral_cube/analysis_utilities.py
```diff
@@ -0,0 +1,310 @@
...
+
+def _fourier_shifter(x, shift, axis):
+    '''
+    Helper function for `~fourier_shift`.
+    '''
+    ftx = np.fft.fft(x, axis=axis)
+    m = np.fft.fftfreq(x.shape[axis])
+    m_shape = [1] * len(x.shape)
+    m_shape[axis] = m.shape[0]
+    m = m.reshape(m_shape)
```
keflavich: I believe this could be rephrased as:
```
slices = [slice(None) if ii == axis else None for ii in range(x.ndim)]
m = m[slices]
```
I'm not sure that's actually any clearer...

spectral_cube/analysis_utilities.py
```diff
@@ -0,0 +1,310 @@
...
+                               " spectral unit ().".format(spec_unit))
+
+        v0 = v0.to(spec_unit)
+
+        if v0 < cube.spectral_axis.min() or v0 > cube.spectral_axis.max():
+            raise ValueError("v0 must be within the range of the spectral "
+                             "axis of the cube.")
+
+    # Calculate the pixel shifts that will be applied.
+    vdiff = np.abs(np.diff(cube.spectral_axis[:2])[0])
```
keflavich: Might be worth adding a check like:
```
vdiff2 = np.abs(np.diff(cube.spectral_axis[-2:])[0])
if vdiff2 != vdiff: # should use a np.testing thing here to make sure "close enough" is good enough?
    raise ValueError("Cannot shift spectra on nonlinear axes")
```

spectral_cube/analysis_utilities.py
```diff
@@ -0,0 +1,310 @@
...
+
+    all_shifted_spectra = []
+
+    if chunk_size == -1:
+        chunk_size = len(xy_posns)
+
+    n_chunks = len(xy_posns[0]) / chunk_size
+
+    # Create chunks of spectra for read-out.
+    for i, chunk in enumerate(get_chunks(len(xy_posns[0]), chunk_size)):
```
keflavich: Progressbar here?  Optionally?

spectral_cube/tests/utilities.py
```diff
@@ -0,0 +1,103 @@
...
+    header = generate_header(pixel_scale, spec_scale, beamfwhm, imshape, v0)
+
+    return fits.PrimaryHDU(data, header)
+
+
+def gaussian(x, amp, mean, sigma):
+    return amp * np.exp(- (x - mean)**2 / (2 * sigma**2))
+
+
+def generate_gaussian_cube(shape=(100, 100, 100), sigma=8., amp=1.,
```
keflavich: 100^3 is a bit big for automated tests; maybe 25^3?

keflavich:
Looks good to me.  Do you want to add that last linear axis check or just call it done?

e-koch:
It's on line 208. `vdiff_tol` sets the maximum relative change the vdiff can have (default set to 0.01).

PR #450, opened 2017-12-05
Fix several major issues related to axis=(1,2) operations

keflavich:
First, and most important, it was possible to unintentionally and unacceptably have full-memory runs initiated when explicitly setting `how='slice'` because of a try/except clause I had added (that was very wrong).

To solve this, I just implemented multidimensional averaging in `mean` and `_reduce_slicewise`.    I added some more tests, too, and the ability to require progressbars

### DIFF
--- a/spectral_cube/utils.py
+++ b/spectral_cube/utils.py
@@ -24,14 +24,15 @@ def warn_slow(function):
 
     @wraps(function)
     def wrapper(self, *args, **kwargs):
-        if self._is_huge and not self.allow_huge_operations:
+        warn_how = ('how' in kwargs and kwargs['how'] == 'cube') or 'how' not in kwargs
+        if self._is_huge and not self.allow_huge_operations and warn_how:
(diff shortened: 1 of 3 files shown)


keflavich:
@e-koch if you have time to give this a quick review, please do.  I'm going to merge it, though, as it includes a critical fix.

spectral_cube/utils.py
```diff
@@ -24,14 +24,16 @@ def warn_slow(function):
 
     @wraps(function)
     def wrapper(self, *args, **kwargs):
-        if self._is_huge and not self.allow_huge_operations:
+        # if the function accepts a 'how', the 'cube' approach requires the whole cube in memory
+        warn_how = ('how' in kwargs and kwargs['how'] == 'cube') or 'how' not in kwargs
```
e-koch: This can be simplified a bit: `warn_how = kwargs.get('how') == 'cube'`

PR #453, opened 2017-12-21
Fix a few minor issues with stacking

e-koch:
A few minor things based on using the stacking function with various datasets. The only change in behaviour is using `cube.filled_data` instead of `cube.unmasked_data`. Presumably most use-cases should be applied with whatever mask has been applied to the cube.


keflavich:
:+1:, should work on masked data.

Issue #452, opened 2017-12-21
Multi-threading compatibility of SpectralCube Smoothing Functions


robelgeda:
Hello, I am working on CubeViz at STScI and we were hoping to use Spectral Cube's smoothing capeblities. The current version of SpectralCube.smoothing functions use [astropy.utils.console.ProgressBar](https://github.com/radio-astro-tools/spectral-cube/blob/cc90374cb3a8efdeb3e5d32731b1178b08b29281/spectral_cube/spectral_cube.py#L2480) which demanded that the smoothing function be executed on the main thread. Since we are using a GUI, this causes significant competition for resources. It would help us a great deal if I can add the option of 'turning off' the `astropy.utils.console.ProgressBar` in the smoothing functions. We are also interested in adding the ability to abort smoothing mid-calculation and updating a pyqt `QProgressBar`. I have made a #454  that would resolve our issues.

keflavich:
I'm happy to see the `ProgressBar` become optional, so please submit your PR.  Note that the link in the original post here points back to this issue (#452), so it appears you have not yet submitted the PR.

robelgeda:
Hi @keflavich! Thank you! I just submitted a [PR](#454).

PR #454, opened 2017-12-21
Multi-threading Compatibility of Smoothing Functions

robelgeda:
Please see #452 for the motivation of this work. 

The current version of SpectralCube smoothing functions use `astropy.utils.console.ProgressBar` which demanded that the smoothing function be executed on the main thread. If this requirement is not met, an exception is raised (`ValueError: signal only works in main thread` ). To resolve this issue this PR aims to provide alternatives to `astropy.utils.console.ProgressBar`. This is achieved by letting the user provide the update function which is responsible for incrementing the progress bar. To be specific, the user sets the `external_update_function` class variable before executing the smoothing functions which in turn disable `astropy.utils.console.ProgressBar` in the smoothing functions.

This PR will also adds the ability to abort smoothing mid calculation. This is achieved by adding a boolean class variable `abort` which will serve as a flag to abort the calculations.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2363,7 +2363,6 @@ class SpectralCube(BaseSpectralCube):
             self.pixels_per_beam = (self.beam.sr /
                                     (astropy.wcs.utils.proj_plane_pixel_area(self.wcs) *
                                      u.deg**2)).to(u.dimensionless_unscaled).value
-        self.external_update_function = None  # External function that updates a gui's progress bar
 
     def _new_cube_with(self, **kwargs):
         beam = kwargs.pop('beam', None)
@@ -2398,7 +2397,7 @@ class SpectralCube(BaseSpectralCube):
 
         return newcube
 
-    def spatial_smooth_median(self, ksize, **kwargs):
+    def spatial_smooth_median(self, ksize, update_function=None, **kwargs):
         """
         Smooth the image in each spatial-spatial plane of the cube using a median filter.
 
@@ -2421,11 +2420,9 @@ class SpectralCube(BaseSpectralCube):
                       self.mask.include(view=(ii, slice(None), slice(None))))
                       for ii in range(self.shape[0]))
 
-        if self.external_update_function is None:
+        if update_function is None:
             pb = ProgressBar(shape[0])
             update_function = pb.update
-        else:
-            update_function = self.external_update_function
 
         def _gsmooth_image(args):
             """
@@ -2457,7 +2454,9 @@ class SpectralCube(BaseSpectralCube):
 
     def spatial_smooth(self, kernel,
                        #numcores=None,
-                       convolve=convolution.convolve, **kwargs):
+                       convolve=convolution.convolve,
+                       update_function=None,
+                       **kwargs):
         """
         Smooth the image in each spatial-spatial plane of the cube.
 
@@ -2482,11 +2481,9 @@ class SpectralCube(BaseSpectralCube):
                      self.mask.include(view=(ii, slice(None), slice(None))))
                      for ii in range(self.shape[0]))
 
-        if self.external_update_function is None:
+        if update_function is None:
             pb = ProgressBar(shape[0])
             update_function = pb.update
-        else:
-            update_function = self.external_update_function
 
         def _gsmooth_image(args):
             """
@@ -2517,7 +2514,7 @@ class SpectralCube(BaseSpectralCube):
 
         return newcube
 
-    def spectral_smooth_median(self, ksize, **kwargs):
+    def spectral_smooth_median(self, ksize, update_function=None, **kwargs):
         """
         Smooth the cube along the spectral dimension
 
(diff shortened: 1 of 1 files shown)


robelgeda:
I should also mention. If progress bar is not needed at all. One can point the class var `external_update_function` to a dummy function which simply returns.

spectral_cube/spectral_cube.py
```diff
@@ -2363,6 +2363,8 @@ def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
             self.pixels_per_beam = (self.beam.sr /
                                     (astropy.wcs.utils.proj_plane_pixel_area(self.wcs) *
                                      u.deg**2)).to(u.dimensionless_unscaled).value
+        self.abort = False  # Checked mid calculation
```
keflavich: It looks like `abort` should be a local variable, not a class instance attribute.  Is there any reason to define this as `self.abort` instead of just `abort`?

spectral_cube/spectral_cube.py
```diff
@@ -2442,6 +2451,8 @@ def _gsmooth_image(args):
                                                            )
                                        )
                                   ])
+        if self.abort:
+            return
```
keflavich: Could we instead return a partial result here?

spectral_cube/spectral_cube.py
```diff
@@ -2363,6 +2363,8 @@ def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
             self.pixels_per_beam = (self.beam.sr /
                                     (astropy.wcs.utils.proj_plane_pixel_area(self.wcs) *
                                      u.deg**2)).to(u.dimensionless_unscaled).value
+        self.abort = False  # Checked mid calculation
+        self.external_update_function = None  # External function that updates a gui's progress bar
```
keflavich: Rather than have this as a class attribute, why not just make it a keyword argument to the function?

spectral_cube/spectral_cube.py
keflavich: I see the `abort_function` below; is the idea that this function would be called and somehow update the state of the `Cube` object *during* the smoothing operation?  I was unaware that that is possible.

spectral_cube/spectral_cube.py
```diff
@@ -2420,14 +2422,21 @@ def spatial_smooth_median(self, ksize, **kwargs):
...
+        else:
+            update_function = self.external_update_function
 
         def _gsmooth_image(args):
             """
             Helper function to smooth a spectrum
             """
             (im, includemask),kwargs = args
-            pb.update()
+            if self.abort:
```
e-koch: The current issue with using multi-processing here is that `_gsmooth_image` isn't picklable. I'm concerned that adding a check for a class attribute might make this issue harder to fix in the future. Thoughts @keflavich?

spectral_cube/spectral_cube.py
robelgeda: Thank you for the quick review. Yes this is exactly it does. During the smoothing operation, a separate thread with an abort button can call `abort_function` which in turn would toggle the value of `self.abort` to true. If we make these variables local to the function, we would not be able to externally change their states.

spectral_cube/spectral_cube.py
robelgeda: Yes we can, where can. Which variable would be appropriate for this?

spectral_cube/spectral_cube.py
robelgeda: If we make these variables local to the function, we would not be able to externally change their states.

robelgeda:
An alternative to the `abort` class variable solution here would be to check in the gui/external update function if an abort condition has been met, which would raise an exception. This would completely remove the `self.abort` checks, function and variable. Would this be a better solution?

robelgeda:
I have updated the PR by removing the abort functionality. For my update_function, if an abort is called, then the update function will raise a custom exception. But the main issue I would like to resolve with this PR is having an alternative for `astropy.utils.console.ProgressBar`.

spectral_cube/spectral_cube.py
```diff
@@ -2363,6 +2363,7 @@ def __init__(self, data, wcs, mask=None, meta=None, fill_value=np.nan,
             self.pixels_per_beam = (self.beam.sr /
                                     (astropy.wcs.utils.proj_plane_pixel_area(self.wcs) *
                                      u.deg**2)).to(u.dimensionless_unscaled).value
+        self.external_update_function = None  # External function that updates a gui's progress bar
```
keflavich: Again, why not make this a function parameter instead of a class attribute?

robelgeda: actually there is no need for that on my end. I can make this a parameter.

robelgeda:
@keflavich: I updated the PR by turning the class variables into parameters.

robelgeda:
I almost forgot CC: @astrofrog

robelgeda:
I have added this update to all functions that use `astropy.utils.console.ProgressBar`. (Not to single out smoothing functions)

robelgeda:
@keflavich @e-koch : I think its ready. Please let know if this works.

robelgeda:
@keflavich Thank you very much! This really is a great help to us.

PR #459, opened 2018-01-25
Let apply_function method return cube-like objects; custom update function

nmearl:
This PR adds the ability to return cube-like objects from the `apply_function` method, and also allows for providing a custom progress update function à la #454.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -761,7 +761,8 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         return self._new_cube_with(data=data, unit=unit)
 
     def apply_function(self, function, axis=None, weights=None, unit=None,
-                       projection=False, progressbar=False, **kwargs):
+                       projection=False, progressbar=False, 
+                       update_function=None, keep_shape=False, **kwargs):
         """
         Apply a function to valid data along the specified axis or to the whole
         cube, optionally using a weight array that is the same shape (or at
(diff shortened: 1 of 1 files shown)


keflavich:
These changes look fine, so I'll merge once we resolve test issues.

However, `apply_function` appears to be narrower in scope than its name applies.  `apply_function`, prior to this PR, could reduce the dimensionality of a cube entirely (to a scalar) or reduce it by 1, to a plane.  It could not return 3D or 1D results.  After this PR, it can return 3D results, but still not 1D.  Perhaps we should expand it in that direction?

spectral_cube/spectral_cube.py
```diff
@@ -785,6 +786,13 @@ def apply_function(self, function, axis=None, weights=None, unit=None,
         progressbar : bool
             Show a progressbar while iterating over the slices/rays through the
             cube?
+        keep_shape : bool
+            If `True`, the returned object will be the same dimensionality as
+            the cube.
+        update_function : function
+            An alternative tracker for the progress of applying the function
+            to the cube data. If `progressbar` is `True`, this argument is
```
keflavich: ``progressbar``

spectral_cube/spectral_cube.py
```diff
@@ -791,7 +791,7 @@ def apply_function(self, function, axis=None, weights=None, unit=None,
             the cube.
         update_function : function
             An alternative tracker for the progress of applying the function
-            to the cube data. If `progressbar` is `True`, this argument is
+            to the cube data. If progressbar is true, this argument is
```
keflavich: You don't like using double-backticks?  I realize my comment before just looked like ``progressbar``; I was trying to show \`\`progressbar\`\`

nmearl: Hah, got it! I was wondering if that's what you meant, but figured there was some sphinx flag that just wouldn't allow formatting in docstrings for whatever reason.

keflavich: Yeah, I think this should work now.  Let's see....

keflavich:
Yeah, I think we've fixed the issue that was keeping CI from completing.  Let's see....

PR #464, opened 2018-01-31
when creating composite masks, allow numpy arrays in the same way that spectral_cube.with_mask would

keflavich:
Currently, if you try something like:

```
cube.with_mask((cube > 5*cube.unit) & (cube.max(axis=0) > 10*cube.unit))
```

you'll get an uninformative `AttributeError: 'numpy.ndarray' object has no attribute '_validate_wcs'` because `with_mask` isn't handling the input mask, CompositeMask is.  This fixes that.

TODO:

 * [x] write a test

### DIFF
--- a/spectral_cube/masks.py
+++ b/spectral_cube/masks.py
@@ -339,7 +339,7 @@ class CompositeMask(MaskBase):
         mask1._validate_wcs(mask2._wcs)
         # In order to composite composites, they must have a _wcs defined.
         # (maybe this should be a property?)
-        self._wcs = self._mask1._wcs
+        self._wcs = mask1._wcs
 
         self._mask1 = mask1
         self._mask2 = mask2
--- a/spectral_cube/masks.py
+++ b/spectral_cube/masks.py
@@ -334,6 +334,13 @@ class CompositeMask(MaskBase):
                                  "%s vs %s" % (mask2.shape, mask1.shape))
             mask2 = BooleanArrayMask(mask2, mask1._wcs, shape=mask1.shape)
 
+        # both entries must have compatible, which effectively means
+        # equal, WCSes.
+        self._mask1._validate_wcs(self._mask2._wcs)
+        # In order to composite composites, they must have a _wcs defined.
+        # (maybe this should be a property?)
+        self._wcs = self._mask1._wcs
+
         self._mask1 = mask1
         self._mask2 = mask2
         self._operation = operation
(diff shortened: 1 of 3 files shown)


keflavich:
@e-koch mind skimming this one?

spectral_cube/masks.py
```diff
@@ -314,13 +323,52 @@ class CompositeMask(MaskBase):
...
+            # In order to composite composites, they must have a _wcs defined.
+            # (maybe this should be a property?)
+            self._wcs = mask1._wcs
+        elif hasattr(mask1, '_wcs'):
+            # if one mask doesn't have a WCS, but the other does, the
+            # compositemask should have the same WCS as the one that does
+            self._wcs = mask1._wcs
+        elif hasattr(mask2, '_wcs'):
+            self._wcs = mask2._wcs
+
```
e-koch: Can the `else` case ever happen? Presumably not if someone is masking via a `SpectralCube`. If for some reason you used `CompositeMask` on its own and gave it two numpy arrays, it wouldn't fail until a `_validate_wcs` call is triggered. We could raise a WCS error just in case.

spectral_cube/tests/test_masks.py
```diff
@@ -543,3 +542,16 @@ def test_filled():
...
+    cube, data = cube_and_raw('adv.fits')
+
+    med = cube.median()
+    mask = cube > med
+    arrmask = cube.max(axis=0) > med
+
+    # we're just testing that this doesn't fail
+    combined_mask = mask & arrmask
+
+    mcube = cube.with_mask(combined_mask)
```
e-koch: Add some sort of operation here that will trigger a call to `mask.include` to make sure the mask has the correct WCS object. I don't think `validate_wcs` will be called in the mask until an operation is performed on the masked cube.

spectral_cube/masks.py
keflavich: If you do something like `cube.with_mask(FunctionMask(lambda x: x>5) & (cube > 5))` (which is redundant...), the first mask would not have a `_wcs` attribute, but the second would

e-koch: I'm referring to the case where neither has a WCS object. It would only happen (I think) if you're creating a `CompositeMask` directly from two numpy arrays. I can't think of _why_ you would ever want to do this, so adding an error in the else case is largely redundant. You're welcome to ignore this point.

keflavich: my bad, you're right.  You could get it from composing two `FunctionMask`s though

keflavich: The fail case here is just to leave `_wcs` undefined, though. I think that's actually the desired behavior since I'm doing `hasattr` tests

e-koch: Sounds good!

PR #465, opened 2018-02-07
Possible fixes to the stacking code

keflavich:
cc @e-koch - the padding *might* be off.  Testing underway...


spectral_cube/analysis_utilities.py
```diff
@@ -269,7 +276,10 @@ def stack_spectra(cube, velocity_surface, v0=None,
 
             all_shifted_spectra.extend([out for out in shifted_spectra])
 
-    stacked = stack_function(all_shifted_spectra, axis=0)
+    shifted_spectra_array = np.array(all_shifted_spectra)
+    assert shifted_spectra_array.ndim == 2
```
keflavich: Just for the record, this is to solve a different issue - I once encountered an empty or weirdly shaped `all_shifted_spectra`, and I want to prevent that.  Obviously assertions are not that useful in production code, but it should make future related errors easier to trace.

PR #466, opened 2018-02-07
Stacker: Do not allow mismatched shapes!

keflavich:
cc @e-koch - you could put in any arbitrary velocity surface map and it would "work" incorrectly.


e-koch:
Good catch!

PR #469, opened 2018-02-23
allow an equivalency to be specified for Jy<->K in LDOs, don't force them to have frequency defined

keflavich:
currently, trying to do `projection.to(u.K, equivalency)` will fail if no frequency is associated with the ldo


keflavich:
I can't reproduce these errors on astropy 2.0.4

PR #474, opened 2018-03-07
Memory mapping in smoothing operations

keflavich:
The previous implementation stored the full cube in memory as a list of spectra without ever checking the cube size (bad!) and duplicated that object in memory a second time when doing `np.array([list of spectra])`.

This is an attempt to work around that using memory maps by default and allocating the full memory otherwise.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,8 @@
 ------------------
  - Refactor spectral smoothing tools to allow parallelized application *and*
    memory mapped output (to avoid loading cube into memory).  Created
-   ``apply_function_parallel_spectral`` to make this general.
+   ``apply_function_parallel_spectral`` to make this general.  Added
+   ``joblib`` as a dependency.
    (https://github.com/radio-astro-tools/spectral-cube/pull/474)
 
 0.4.2 (2018-02-21)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,9 @@
 0.4.3 (unreleased)
 ------------------
- - none yet
+ - Refactor spectral smoothing tools to allow parallelized application *and*
+   memory mapped output (to avoid loading cube into memory).  Created
+   ``apply_function_parallel_spectral`` to make this general.
+   (https://github.com/radio-astro-tools/spectral-cube/pull/474)
 
 0.4.2 (2018-02-21)
 ------------------
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -10,6 +10,7 @@ import operator
 import re
 import itertools
 import copy
+import tempfile
 
 import astropy.wcs
 from astropy import units as u
@@ -2639,6 +2640,7 @@ class SpectralCube(BaseSpectralCube):
                         #numcores=None,
                         convolve=convolution.convolve,
                         update_function=None,
+                        use_memmap=True,
                         **kwargs):
         """
         Smooth the cube along the spectral dimension
@@ -2654,6 +2656,9 @@ class SpectralCube(BaseSpectralCube):
         update_function : method
             Method that is called to update an external progressbar
             If provided, it disables the default `astropy.utils.console.ProgressBar`
+        use_memmap : bool
+            If specified, a memory mapped temporary file on disk will be
+            written to rather than storing the intermediate spectra in memory.
         kwargs : dict
             Passed to the convolve function
         """
@@ -2663,8 +2668,10 @@ class SpectralCube(BaseSpectralCube):
         # "cubelist" is a generator
         # the boolean check will skip smoothing for bad spectra
         # TODO: should spatial good/bad be cached?
-        cubelist = ((self.filled_data[:,jj,ii],
-                     self.mask.include(view=(slice(None), jj, ii)))
+        spectra = ((self.filled_data[:,jj,ii],
+                    self.mask.include(view=(slice(None), jj, ii)),
+                    ii, jj,
+                   )
                     for jj in range(self.shape[1])
                     for ii in range(self.shape[2]))
 
@@ -2672,33 +2679,27 @@ class SpectralCube(BaseSpectralCube):
             pb = ProgressBar(shape[1] * shape[2])
             update_function = pb.update
 
+        if use_memmap:
+            ntf = tempfile.NamedTemporaryFile()
+            smoothcube = np.memmap(ntf, mode='w+', shape=shape, dtype=np.float)
+        else:
+            smoothcube = np.empty(shape=shape, dtype=np.float)
+
         def _gsmooth_spectrum(args):
             """
             Helper function to smooth a spectrum
             """
-            (spec, includemask),kernel,kwargs = args
+            (spec, includemask, ii, jj),kernel,kwargs = args
             update_function()
 
             if any(includemask):
-                return convolve(spec, kernel, normalize_kernel=True, **kwargs)
+                smoothcube[:,jj,ii] = convolve(spec, kernel, normalize_kernel=True, **kwargs)
             else:
-                return spec
+                smoothcube[:,jj,ii] = spec
 
         # could be numcores, except _gsmooth_spectrum is unpicklable
         with cube_utils._map_context(1) as map:
-            smoothcube_ = np.array([x for x in
-                                    map(_gsmooth_spectrum, zip(cubelist,
-                                                               itertools.cycle([kernel]),
-                                                               itertools.cycle([kwargs]),
-                                                              )
-                                       )
-                                   ]
-                                  )
-
-        # empirical: need to swapaxes to get shape right
-        # cube = np.arange(6*5*4).reshape([4,5,6]).swapaxes(0,2)
-        # cubelist.T.reshape(cube.shape) == cube
-        smoothcube = smoothcube_.T.reshape(shape)
+            map(_gsmooth_spectrum, zip(spectra, itertools.cycle([kernel]), itertools.cycle([kwargs]),))
 
         # TODO: do something about the mask?
         newcube = self._new_cube_with(data=smoothcube, wcs=self.wcs,
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2608,12 +2608,12 @@ class SpectralCube(BaseSpectralCube):
         if not scipyOK:
             raise ImportError("Scipy could not be imported: this function won't work.")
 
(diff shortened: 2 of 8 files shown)


keflavich:
@astrofrog do you have any idea how to run a numpy operation in parallel across an array?  It turns out to be really hard, I believe, because `multiprocessing` passes values by pickle, which means there is no way to call multiple copies of the same function with the same array (same memory object) in different processes.

This is really annoying because operations like `spectral_smooth` are trivial to parallelize *if* you can run the commands in parallel, but it is generally not possible or desirable to have all of the data to be operated on in memory simultaneously.  What we really want to do is: read 10^-6 of the cube, operate on it, write that back out, and repeat that process (in parallel) 10^6 times.

keflavich:
I've settled on `joblib` as the right approach, since it has [an example](https://pythonhosted.org/joblib/parallel.html) built *exactly* like the use case I already defined, even including numpy memmap arrays.  It works beautifully for our tiny, simple test case.

Unfortunately, there's still more work to do in the case where we care about what the operation does to the mask.  I know I can do it, but it will take some time and thinking and a lot more tests...
...but once we have that in place, we can do everything in parallel!  Like statcont, etc.

keflavich:
This has turned into a big and, imo, awesome change.  I'm requesting some reviews.

keflavich:
In particular, let me know if you have ideas on how to test, e.g., that cubes aren't loaded into memory when using memmap mode, or maybe speed tests.

spectral_cube/spectral_cube.py
```diff
@@ -2580,69 +2585,123 @@ def spectral_smooth_median(self, ksize, update_function=None, **kwargs):
...
+        update_function : method
+            Method that is called to update an external progressbar
+            If provided, it disables the default `astropy.utils.console.ProgressBar`
+        use_memmap : bool
+            If specified, a memory mapped temporary file on disk will be
+            written to rather than storing the intermediate spectra in memory.
+        num_cores : int or None
+            The number of cores to use if running in parallel
+        kwargs : dict
+            Passed to the convolve function
```
e-koch: Remove convolve

spectral_cube/tests/test_regrid.py
```diff
@@ -114,6 +114,34 @@ def test_spectral_smooth():
...
+    cube, data = cube_and_raw('522_delta.fits')
+
+    result = cube.spectral_smooth(kernel=convolution.Gaussian1DKernel(1.0), num_cores=4)
+
+    np.testing.assert_almost_equal(result[:,0,0].value,
+                                   convolution.Gaussian1DKernel(1.0,
+                                                                x_size=5).array,
+                                   4)
+
+    result = cube.spectral_smooth(kernel=convolution.Gaussian1DKernel(1.0), num_cores=4, use_memmap=True)
```
e-koch: I think `use_memmap=False` here. The default is True above.

CHANGES.rst
```diff
@@ -1,6 +1,9 @@
 0.4.3 (unreleased)
 ------------------
- - none yet
+ - Refactor spectral smoothing tools to allow parallelized application *and*
+   memory mapped output (to avoid loading cube into memory).  Created
+   ``apply_function_parallel_spectral`` to make this general.
+   (https://github.com/radio-astro-tools/spectral-cube/pull/474)
```
e-koch: Note here if we're introducing joblib as a dependency.

spectral_cube/spectral_cube.py
```diff
@@ -2580,69 +2585,123 @@ def spectral_smooth_median(self, ksize, update_function=None, **kwargs):
...
-                                       )
-                                   ]
-                                  )
-
-        # empirical: need to swapaxes to get shape right
-        # cube = np.arange(6*5*4).reshape([4,5,6]).swapaxes(0,2)
-        # cubelist.T.reshape(cube.shape) == cube
-        smoothcube = smoothcube_.T.reshape(shape)
+        # TEST
+        from joblib import Parallel, delayed
```
e-koch: Will the map context manager eventually be used here?

spectral_cube/tests/test_regrid.py
```diff
@@ -114,6 +114,34 @@ def test_spectral_smooth():
...
+    result = cube.spectral_smooth(kernel=convolution.Gaussian1DKernel(1.0), use_memmap=True)
+
+    np.testing.assert_almost_equal(result[:,0,0].value,
+                                   convolution.Gaussian1DKernel(1.0,
+                                                                x_size=5).array,
+                                   4)
+
+
+# TODO: uncomment this when we figure out how to make it work
+def test_spectral_smooth_4cores():
```
e-koch: `skipif` check for joblib

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -1680,6 +1680,17 @@ def test_spectral_smooth_median():
 
     np.testing.assert_almost_equal(cube_spectral_median[:,1,1].value, result)
 
+@pytest.mark.skipif('not SCIPYOK')
```
e-koch: `skipif` check for joblib

CHANGES.rst
keflavich: I'd prefer to add it to the REQUIREMENTS so that it gets pip installed w/spectral-cube

spectral_cube/spectral_cube.py
keflavich: Actually, no, my plan is to factor out the map context manager entirely; it was a failure.  multiprocess pool mapping didn't work for any of the intended use cases.

keflavich:
I haven't had a chance to test yet.  That's a disappointing performance improvement floor.

I don't believe the generator should load the whole thing in; if I understand generators correctly, no operation will be performed until the particular iteration is called, e.g.:
```
In [4]: def blah(x):
    print(x)
    return x
   ...:

In [5]: z = (blah(x) for x in range(10))

In [6]: list(z)
0
1
2
3
4
5
6
7
8
9
Out[6]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

keflavich:
@e-koch could you post your test code here?  It would be convenient to reuse it (and add it somewhere in the test codebase...)

keflavich:
Well... I wrote up some performance tests, and parallelism makes it worse for a small cube (100x25x25, 300x25x25).  I see a small benefit when moving to a bigger one (300x100x100):
``` 
# keys are # of cores
{1: 18.52146705100313, 2: 13.401284503983334, 3: 13.308872838970274, 4: 13.208968839026056}
```

keflavich:
...why is >2 cores the same as 2 cores?

keflavich:
In my latest test, it looks like njobs>2 doesn't use >2 cores.

e-koch:
Here's the code I was testing with:
```
kern = convolution.Gaussian1D(5)
times = []
for n in range(1,7):
   start = time.time()
   smooth = cube.spectral_smooth(kern, num_cores=n, use_memmap=True)
   end = time.time()
   times.append(end - start)
```
I'm getting the correct number of processes spawned for `njobs>2`, but they don't seem to actually process anything.

keflavich:
according to my machine, only a single cpu is being used when I use n_jobs > 1.

keflavich:
how are you testing how many jobs are spawned?  I was just watching the CPU usage in activity monitor, and it only went up to 100% in a single process.

e-koch:
Just using `top` in a terminal. The right number of processes are spawned for me, but they aren't being used.

keflavich:
dang.  I don't know what's going wrong.  joblib docs aren't super helpful.

CHANGES.rst
e-koch: Agreed. I thought it would be good to note where it was introduced here.

keflavich:
I was wrong, joblib isn't doing anything at all, and turns out to maybe be as useless as multiprocessing.  Apparently in order to parallelize a function, that function needs to be pickleable, and in order to be pickleable, an object needs to be defined at the top level of the source code.  This is, by definition, not what we want (we only want to parallelize dynamically defined functions).

keflavich:
I may have found a way around the above problem, but it means we have to drop the progressbar and make other code a bit uglier.

keflavich:
even when running in parallel, the net effect seems to be worse performance. =(

keflavich:
Maybe a little better for 500x64x64?

```
{1: 10.023035926977172, 2: 8.497304802993312, 3: 6.610799013986252, 4: 6.596845873980783}
```

and for 1000x64x64, it's actually OK:
```
memmap=True
{1: 17.087035791017115, 2: 12.494161848095246, 3: 9.698949208017439, 4: 9.414709302014671}
```

So maybe this is a genuine use case that only works in the large-N limit?  Neat?

keflavich:
eh... still annoyingly inefficient:
```
memmap=True
{1: 117.6646711260546, 2: 61.74993083393201, 3: 45.986659401096404, 4: 42.20888108399231}
```

keflavich:
I'm on a 4-core machine with pseudo-8 cores with hyperthreading.  I don't know why 3 behaves like 4, but whatever....

```
memmap=True shape=(300, 64, 64)
{0: 6.062829509028234, 1: 7.174146079923958, 2: 7.156198110082187, 3: 6.10964499192778, 4: 5.964741178089753}

memmap=True shape=(600, 64, 64)
{0: 10.829934995970689, 1: 11.859699308988638, 2: 10.380743535002694, 3: 9.24959541298449, 4: 9.427205897984095}

memmap=True shape=(900, 64, 64)
{0: 14.884416320011951, 1: 16.30231760093011, 2: 11.62743233202491, 3: 9.293404301046394, 4: 8.565535836969502}

memmap=True shape=(300, 128, 128)
{0: 23.03774840501137, 1: 27.81474193208851, 2: 15.414176015066914, 3: 12.118889170931652, 4: 12.681411720928736}
```
(more running)

keflavich:
just to confirm, hyperthreading = no improvement
```
memmap=True shape=(300, 128, 128)
{0: 23.12492688500788, 1: 28.01206497394014, 2: 16.355690013966523, 3: 12.66177305392921, 4: 13.398187862010673, 5: 14.465838123927824, 6: 14.099746330059133, 7: 14.005758916959167, 8: 13.764693430042826}
```

keflavich:
I plan to merge this later today.  Please submit bug reports if you come up with anything!  This is a pretty substantial change, and I'll need to do the same thing for spatial smoothing.  I also plan to extend this to applying other interesting astropy functions, such as sigma clipping (to incorporate `statcont` into spectral-cube)

PR #477, opened 2018-03-28
bottleneck.nanmedian does not allow axis with a tuple/list.

### LINKED ISSUE #476: `bottleneck.nanmedian` does not allow axis to have a tuple/list
The current version of `bottleneck.nanmedian` (v1.2.1) doesn't allow axis to have a tuple/list, for example:
```
import numpy as np
import bottleneck as bnok

A = np.random.normal(0, 100, (20, 20, 500))

print(np.nanmedian(A, axis=(1, 2)))
print(bnok.nanmedian(A, axis=(1, 2)))
```

This produces an error:
```
TypeError: `axis` must be an integer or None
```

Then, if bottleneck is in the environment I get fails in the `cube.median()` code if I pass in axis as a list or tuple. 

I have a fix and will create a PR.

brechmos-stsci:
As stated, as of version 1.2.1 of bottleneck it does not allow a tuple/list for the axis.  So, this adds a check to see if that is the case.

Fixes #476

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -974,7 +974,8 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
             bnok = False
 
         # slicewise median is nonsense, must force how = 'cube'
-        if bnok and not iterate_rays:
+        # bottleneck.nanmedian does not allow axis to be a list or tuple
+        if bnok and not iterate_rays and not isinstance(axis, (list, tuple)):
             log.debug("Using bottleneck nanmedian")
             result = self.apply_numpy_function(nanmedian, axis=axis,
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@keflavich If you are good with this fix, it would be great to commit (fixes some spectral stuff I am working on). If you have a better suggestion for a fix, please feel free to post and I'll correct.

keflavich:
Looks good, thanks for the catch.  It's annoying that they don't support something so basic.

Could you add a test that catches this?

brechmos-stsci:
@keflavich I added the test and seems to have passed (wahoo!).

keflavich:
@astrofrog has a workaround:
https://github.com/glue-viz/glue/pull/1641/files#diff-4e6dc6cf25762999955424dd2d995973R300

PR #478, opened 2018-03-28
MAJOR: WCS reversal is non-functional

keflavich:
If you slice a cube with [::-1], it does not return a valid cube.  This is a
severe error that I don't really comprehend yet, so I'm starting with the
regression test.

### DIFF
--- a/spectral_cube/tests/test_spectral_cube.py
+++ b/spectral_cube/tests/test_spectral_cube.py
@@ -957,6 +957,11 @@ def test_slice_wcs_reversal():
                                   cube.spectral_axis.value[::-1])
     np.testing.assert_array_equal(rrcube.world_extrema.value,
                                   cube.world_extrema.value)
+    # check that the lon, lat arrays are *entirely* unchanged
+    np.testing.assert_array_equal(rrcube.spatial_coordinate_map[0].value,
+                                  cube.spatial_coordinate_map[0].value)
+    np.testing.assert_array_equal(rrcube.spatial_coordinate_map[1].value,
+                                  cube.spatial_coordinate_map[1].value)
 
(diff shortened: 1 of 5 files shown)


keflavich:
@astrofrog - does this ring any bells?  Somehow our tests didn't cover this, which is a HUGE oversight.

keflavich:
I cancelled some of the travis builds since they were intentional failures

keflavich:
several of the travis jobs have failed on code that is *not present* in the latest commit.  This appears to be a travis bug?  https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/359491426#L1246

re-running travis...

keflavich:
@astrofrog I think this turned out to be a very simple oversight, where I simply forgot to set one of the WCS parameters.  Could you just verify that I've done something halfway sensible, with the tests if nothing else?  This is an urgent bugfix, but I don't want to merge it without someone else's eyes on it.

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -940,6 +941,21 @@ def test_slice_wcs(view, naxis):
...
+    rcube = cube[view]
+    rrcube = rcube[view]
+
+    np.testing.assert_array_equal(np.diff(cube.spectral_axis),
+                                  -np.diff(rcube.spectral_axis))
+
+    np.testing.assert_array_equal(rrcube.spectral_axis.value,
+                                  cube.spectral_axis.value)
+    np.testing.assert_array_equal(rcube.spectral_axis.value,
+                                  cube.spectral_axis.value[::-1])
```
astrofrog: Can you check that the world coordinate axes stay the same?

PR #479, opened 2018-04-03
Allow 2D standard deviation to be performed slicewise

keflavich:


keflavich:
well, coveralls has failed.  This is definitely an *increase* in coverage, not a decrease.  It's just not covering more lines....

PR #481, opened 2018-04-11
Strip PC00x00x from headers

keflavich:
Another corner case I hadn't seen before


keflavich:
merge despite travis errors.  they're irrelevant

PR #488, opened 2018-06-14
WIP : replacing pyregions with regions.

sushobhana:
work in progress PR

### DIFF
--- a/spectral_cube/spectral_axis.py
+++ b/spectral_cube/spectral_axis.py
@@ -170,11 +170,10 @@ def doppler_z(restfreq):
 def doppler_beta(restfreq):
     restfreq = restfreq.to_value("GHz")
     return [(u.GHz, u.km / u.s,
-            lambda x: constants.si.c.to_value('km/s') * ((1 - ((x / restfreq) ** 2))
-                       / (1 + ((x / restfreq) ** 2))),
-            lambda x: restfreq * np.sqrt(
-                ((constants.si.c.to_value("km/s") - x)
-                             / (x + constants.si.c.to_value("km/s"))))
+            lambda x: constants.si.c.to_value('km/s') * ((1 - ((x / restfreq) ** 2)) /
+                                                         (1 + ((x / restfreq) ** 2))),
+            lambda x: restfreq * np.sqrt((constants.si.c.to_value("km/s") - x) /
+                                         (x + constants.si.c.to_value("km/s")))
              )]
 
 
@@ -182,9 +181,9 @@ def doppler_gamma(restfreq):
     restfreq = restfreq.to_value("GHz")
     return [(u.GHz, u.km / u.s,
             lambda x: constants.si.c.to_value("km/s") * ((1 + (x / restfreq) ** 2) /
-                    (2 * x / restfreq)),
+                                                         (2 * x / restfreq)),
             lambda x: restfreq * (x / constants.si.c.to_value("km/s") +
-                    np.sqrt((x / constants.si.c.to_value("km/s")) ** 2 - 1))
+                                  np.sqrt((x / constants.si.c.to_value("km/s")) ** 2 - 1))
              )]
 
 
--- a/spectral_cube/spectral_axis.py
+++ b/spectral_cube/spectral_axis.py
@@ -1,5 +1,7 @@
 from __future__ import print_function, absolute_import, division
 
+import numpy as np
+
 from astropy import wcs
 from astropy import units as u
 from astropy import constants
@@ -158,27 +160,32 @@ def get_rest_value_from_wcs(mywcs):
 
 
 def doppler_z(restfreq):
+    restfreq = restfreq.to_value("GHz")
     return [(u.GHz, u.km / u.s,
-                lambda x: (restfreq - x) / x,
-                lambda x: restfreq / (1 + x))]
+            lambda x: (restfreq - x) / x,
+            lambda x: restfreq / (1 + x)
+             )]
 
 
 def doppler_beta(restfreq):
+    restfreq = restfreq.to_value("GHz")
     return [(u.GHz, u.km / u.s,
-            lambda x: si.c.to_value('km/s') * (1 - (x / restfreq) ** 2)
-                    / (1 + (x / restfreq) ** 2),
-            lambda x: np.sqrt(
-                (restfreq * ((si.c.to_value("km/s") - x) / (x + si.c.to_value("km/s")))))
-            )]
+            lambda x: constants.si.c.to_value('km/s') * ((1 - ((x / restfreq) ** 2))
+                       / (1 + ((x / restfreq) ** 2))),
+            lambda x: restfreq * np.sqrt(
+                ((constants.si.c.to_value("km/s") - x)
+                             / (x + constants.si.c.to_value("km/s"))))
+             )]
 
 
 def doppler_gamma(restfreq):
+    restfreq = restfreq.to_value("GHz")
     return [(u.GHz, u.km / u.s,
-                lambda x: si.c.to_value("km/s") * (1 + (x / restfreq)**2) /
-                            (2 * x / restfreq),
-                lambda x: restfreq * ((x / si.c.to_value("km/s") +
-                            np.sqrt((x / si.c.to_value("km/s")) ** 2 - 1)))
-                )]
+            lambda x: constants.si.c.to_value("km/s") * ((1 + (x / restfreq) ** 2) /
+                    (2 * x / restfreq)),
+            lambda x: restfreq * (x / constants.si.c.to_value("km/s") +
+                    np.sqrt((x / constants.si.c.to_value("km/s")) ** 2 - 1))
+             )]
 
 
 def convert_spectral_axis(mywcs, outunit, out_ctype, rest_value=None):
--- a/spectral_cube/spectral_axis.py
+++ b/spectral_cube/spectral_axis.py
@@ -153,6 +153,34 @@ def get_rest_value_from_wcs(mywcs):
         ref_value = mywcs.wcs.restwav*u.m
         return ref_value
 
+# Velocity/frequency equivalencies that are not present in astropy core.
+# Ref: https://casa.nrao.edu/casadocs/casa-5-1.2/reference-material/spectral-frames
+
+
+def doppler_z(restfreq):
+    return [(u.GHz, u.km / u.s,
+                lambda x: (restfreq - x) / x,
+                lambda x: restfreq / (1 + x))]
+
+
+def doppler_beta(restfreq):
+    return [(u.GHz, u.km / u.s,
+            lambda x: si.c.to_value('km/s') * (1 - (x / restfreq) ** 2)
+                    / (1 + (x / restfreq) ** 2),
+            lambda x: np.sqrt(
+                (restfreq * ((si.c.to_value("km/s") - x) / (x + si.c.to_value("km/s")))))
+            )]
+
+
+def doppler_gamma(restfreq):
+    return [(u.GHz, u.km / u.s,
+                lambda x: si.c.to_value("km/s") * (1 + (x / restfreq)**2) /
+                            (2 * x / restfreq),
+                lambda x: restfreq * ((x / si.c.to_value("km/s") +
+                            np.sqrt((x / si.c.to_value("km/s")) ** 2 - 1)))
+                )]
+
+
 def convert_spectral_axis(mywcs, outunit, out_ctype, rest_value=None):
     """
     Convert a spectral axis from its unit to a specified out unit with a given output
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -47,8 +47,8 @@ from .utils import (cached, warn_slow, VarianceWarning, BeamAverageWarning,
                     NotImplementedWarning, SliceWarning, SmoothingWarning,
                     StokesWarning, ExperimentalImplementationWarning,
                     BeamAverageWarning, NonFiniteBeamsWarning)
-from .spectral_axis import determine_vconv_from_ctype, get_rest_value_from_wcs, \
-                    doppler_beta, doppler_gamma, doppler_z
+from .spectral_axis import (determine_vconv_from_ctype, get_rest_value_from_wcs,
+                            doppler_beta, doppler_gamma, doppler_z)
 
 from distutils.version import LooseVersion
 
@@ -1800,58 +1800,58 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
 
         return self[slices]
 
-    def subcube_from_ds9region(self, ds9region, allow_empty=False):
+    def subcube_from_ds9region(self, ds9_region, allow_empty=False):
         """
         Extract a masked subcube from a ds9 region
         (only functions on celestial dimensions)
 
         Parameters
         ----------
-        ds9region: str
-            The region to extract
+        ds9_region: str
+            The DS9 region(s) to extract
         allow_empty: bool
             If this is False, an exception will be raised if the region
             contains no overlap with the cube
         """
         import regions
 
-        if isinstance(ds9region, six.string_types):
-            ds9region = regions.DS9Parser(ds9region).shapes.to_regions()
+        if isinstance(ds9_region, six.string_types):
+            region_list = regions.DS9Parser(ds9_region).shapes.to_regions()
         else:
-            raise TypeError("{0} should be a DS9 string".format(ds9region))
+            raise TypeError("{0} should be a DS9 string".format(ds9_region))
 
-        return self.subcube_from_regions(ds9region, allow_empty)
+        return self.subcube_from_regions(region_list, allow_empty)
 
-    def subcube_from_crtf(self, region, allow_empty=False):
+    def subcube_from_crtf(self, crtf_region, allow_empty=False):
         """
         Extract a masked subcube from a CRTF region.
 
         Parameters
         ----------
-        region: str
-            The CRTF region string to extract
+        crtf_region: str
+            The CRTF region(s) string to extract
         allow_empty: bool
             If this is False, an exception will be raised if the region
             contains no overlap with the cube
         """
         import regions
 
-        if isinstance(region, six.string_types):
-            region = regions.CRTFParser(region).shapes.to_regions()
+        if isinstance(crtf_region, six.string_types):
+            region_list = regions.CRTFParser(crtf_region).shapes.to_regions()
         else:
-            raise TypeError("{0} should be a CRTF string".format(region))
+            raise TypeError("{0} should be a CRTF string".format(crtf_region))
 
-        return self.subcube_from_regions(region, allow_empty)
+        return self.subcube_from_regions(region_list, allow_empty)
 
     def subcube_from_regions(self, region_list, allow_empty=False):
         """
-        Extract a masked subcube from a list of `regions.Region` object
+        Extract a masked subcube from a list of ``regions.Region`` object
         (only functions on celestial dimensions)
 
         Parameters
         ----------
-        region_list: `regions.Region` list
-            The region to extract
+        region_list: ``regions.Region`` list
+            The region(s) to extract
         allow_empty: bool, optional
             If this is False, an exception will be raised if the region
             contains no overlap with the cube. Default is False.
@@ -1869,12 +1869,12 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
                 raise TypeError("'{}' should be `regions.Region` object".format(x))
 
         # List of regions are converted to a `regions.CompoundPixelRegion` object.
-        compound_region = _compound_region(regs)
+        compound_region = _regionlist_to_single_region(regs)
 
         # Compound mask of all the regions.
         mask = compound_region.to_mask()
 
-        # Collecting frequency/velocity range, type and rest frequency
+        # Collecting frequency/velocity range, velocity type and rest frequency
         # of each region.
         ranges = [x.meta.get('range', None) for x in regs]
         veltypes = [x.meta.get('veltype', None) for x in regs]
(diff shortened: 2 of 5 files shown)


sushobhana:
@keflavich Can you please look into the code and comment on whether I am going right. ??

spectral_cube/spectral_cube.py
```diff
@@ -3636,3 +3698,12 @@ def mask_channels(self, goodchannels):
 
         return self._new_cube_with(mask=mask,
                                    goodbeams_mask=goodchannels & self._goodbeams_mask)
+
+
+def _compound_region(region_list):
```
keflavich: compound regions belong in the `regions` package rather than here; is your intent to move it there eventually?

sushobhana: ya..it is just for now.

spectral_cube/spectral_cube.py
```diff
@@ -1912,6 +1912,68 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
...
+        mask = compound_region.to_mask()
+
+        xlo, xhi, ylo, yhi = mask.bbox.ixmin, mask.bbox.ixmax, mask.bbox.iymin, mask.bbox.iymax
+
+        # Negative indices will do bad things, like wrap around the cube
+        # If xhi/yhi are negative, there is not overlap
+        if (xhi < 0) or (yhi < 0):
+            raise ValueError("Region is outside of cube.")
+
+        if xlo < 0:
```
keflavich: Just so you're aware: the reason we have this check on the negative side is that indexing with negative numbers does weird and incorrect things.

spectral_cube/spectral_cube.py
```diff
@@ -1912,6 +1912,68 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
...
+
+        if any(dim == 0 for dim in subcube.shape):
+            if allow_empty:
+                warnings.warn("The derived subset is empty: the region does not"
+                              " overlap with the cube (but allow_empty=True).")
+            else:
+                raise ValueError("The derived subset is empty: the region does not"
+                                 " overlap with the cube.")
+
+        maskarray = mask.data[(mask.data.shape[0]-subcube.shape[1]):, (mask.data.shape[1]-subcube.shape[2]):]
```
keflavich: This looks approximately right to me.  Add some comments here, though, to indicate what you're doing so we understand when we read this code later:  you are cropping the mask from the bottom-left because you have just done the same thing with the cube.

We probably also need to crop on the top-right, though, otherwise there's potential for the mask and cube shapes to be mismatched, no?

sushobhana: the test that I added is failing. Can you look into that?? I am not able to understand the cause for failure.

keflavich: could you post the traceback?

sushobhana: ________________________________________________________________ test_ds9region_255_new[255-pixel.reg] ___________________________________________________________________

regfile = '255-pixel.reg'

    @pytest.mark.parametrize('regfile',
                             ('255-fk5.reg', '255-pixel.reg'),
                            )
    def test_ds9region_255_new(regfile):
        # specific test for correctness
        cube, data = cube_and_raw('255.fits')
    
        shapelist = regions.read_ds9(path(regfile))
    
>       subcube = cube.subcube_from_ds9region_new(shapelist)

spectral_cube/tests/test_subcubes.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
spectral_cube/spectral_cube.py:1975: in subcube_from_ds9region_new
    return masked_subcube.minimal_subcube(spatial_only=True)
spectral_cube/spectral_cube.py:1667: in minimal_subcube
    spatial_only=spatial_only)]
spectral_cube/spectral_cube.py:1706: in subcube_slices_from_mask
spectral_cube/masks.py:113: in include
    return self._include(data=data, wcs=wcs, view=view)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <spectral_cube.masks.CompositeMask object at 0x7f3538227e10>
data = array([[[11, 12, 13, 14],
        [16, 17, 18, 19],
        [21, 22, 23, 24]],

       [[36, 37, 38, 39],
        [41, 42, 43, 44],
        [46, 47, 48, 49]]])
wcs = WCS Keywords

Number of WCS axes: 3
CTYPE : 'RA---SIN'  'DEC--SIN'  'VOPT'  
CRVAL : 23.1837500515  30.5765277962  -32...0  
PC3_1 PC3_2 PC3_3  : 0.0  0.0  1.0  
CDELT : -0.000555555561268  0.000555555561268  1288.21496879  
NAXIS : 5  5  2
view = ()

    def _include(self, data=None, wcs=None, view=()):
        result_mask_1 = self._mask1._include(data=data, wcs=wcs, view=view)
        result_mask_2 = self._mask2._include(data=data, wcs=wcs, view=view)
        if self._operation == 'and':
>           return result_mask_1 & result_mask_2
E           TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

spectral_cube/masks.py:377: TypeError
--------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------
[[0. 1. 1. 0.]
 [0. 1. 1. 0.]
 [0. 0. 0. 0.]]
(4, 4) (3, 4)

sushobhana: it is throwing this type error ,
`TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`

keflavich: It looks like the `maskarray` needs to be converted to a boolean array...

keflavich: so, on this line, do `maskarray = (...).astype('bool')`

sushobhana: Thanks. also,I will add some comments and wrt the cropping I will think a little bit more and come back to you .

spectral_cube/spectral_cube.py
sushobhana: yes, it was mentioned in the comments there. Also since the pixel origin (0, 0) is at top left , it makes no sense to have negative pixels. Am I right??

keflavich: That's right.  Except, I regard 0,0 as bottom-left, which is the "image" standard and the one used in astronomy.  0,0 being top-left is the "matrix" standard, which is generally not used in our field.

spectral_cube/tests/test_subcubes.py
```diff
@@ -90,7 +92,43 @@ def test_ds9region_255(regfile):
...
+            sc = cube.subcube_from_ds9region_new(regionlist)
+        # this assertion is redundant, I think...
+        assert exc.errisinstance(result)
+    else:
+        sc = cube.subcube_from_ds9region_new(regionlist)
+        scsum = sc.sum()
+        dsum = data[result].sum()
+        assert_allclose(scsum, dsum)
 
     #region = 'fk5\ncircle(29.9346557, 24.0623827, 0.11111)'
```
sushobhana: I also think there is some problem, because the bounding for a sky region in pyregions is not the same as that of in regions. 
I think there is some problem with the conversion system to pixel.

sushobhana:
Test failings in CI now are not related to this PR, it seems

sushobhana:
@keflavich Any comments on the recent improvements?

spectral_cube/tests/test_subcubes.py
```diff
@@ -150,3 +150,17 @@ def test_ds9region_new(regfile, result):
     #region = 'circle(2,2,2)'
     #subcube = cube.subcube_from_ds9region(region)
 
+
+@pytest.mark.skipif('not regionsOK', reason='Could not import regions')
+def test_regions():
+    cube, data = cube_and_raw('adv.fits')
+
+    # content of image.reg
+    regpix = regions.RectanglePixelRegion(regions.PixCoord(0.5, 1), width=4, height=2)
```
keflavich: This rectangle goes outside the image, which is OK but not great.  `adv.fits` is pretty small, so you want a small region.  This region will go from xlo,xhi,ylo,yhi = [-1.5, 2.5, 0, 2] = [0.5-4/2, 0.5+4/2, 1-2/2, 2+2/2].  So try to pick a pixel coordinate and width/height that fit in the cube better.

keflavich: BUT, it would also be good to have at least three tests: one where the region is in the cube, one where it partly overlaps with the cube, and one where it's outside the cube and therefore fails.

spectral_cube/tests/test_subcubes.py
```diff
@@ -150,3 +150,17 @@ def test_ds9region_new(regfile, result):
...
+
+    regpix.meta['range'] = [-318 * u.km/u.s, -320 * u.km/u.s]
```
keflavich: This looks fine, just make sure you confirm it's in range (I don't remember the velocity range of `adv.fits`...)

spectral_cube/tests/test_subcubes.py
```diff
@@ -150,3 +150,17 @@ def test_ds9region_new(regfile, result):
...
+
+    sc = cube.subcube_from_regions([regpix])
+    scsum = sc.sum()
+    dsum = data[1:-1, 1, :].sum()
+    assert_allclose(scsum, dsum)
```
keflavich: This looks good.  You might also add a test that `cube.mask.include().sum() == [the correct integer value]`

spectral_cube/tests/test_subcubes.py
sushobhana: I have checked the range and set the values accordingly.

spectral_cube/tests/test_subcubes.py
sushobhana: Ah! OK.

spectral_cube/tests/test_subcubes.py
sushobhana: this test is just for checking the extraction of the spectral part. the tests above checks for the spatial one.But, if you want I will do it.

keflavich: OK, that's fine, but the region selection is happening both spatially and spectrally.  As long as you've done the by-hand math to check that the indices below match the selections from the region, it's OK.    You should label this test `_spectral` if it's intended to test only spectral

sushobhana: ya, I better put that `_spectral` label!

spectral_cube/spectral_cube.py
```diff
@@ -1861,7 +1861,7 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
             ylo = 0
 
         log.debug("Region boundaries: ")
-        log.debug("xlo={xlo}, ylo={ylo}, xhi={xhi}, yhi={yhi}".format(xlo=xlo,
+        print("xlo={xlo}, ylo={ylo}, xhi={xhi}, yhi={yhi}".format(xlo=xlo,
```
keflavich: Instead of turning this into a print statement, you can do:
```
from astropy import log
log.setLevel('DEBUG')
```
to enable debug statement prints

spectral_cube/spectral_cube.py
```diff
@@ -1912,6 +1912,122 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
...
+            ds9region = regions.DS9Parser(ds9region).shapes.to_regions()
+        else:
+            raise TypeError("{0} should be a DS9 string".format(ds9region))
+
+        return self.subcube_from_regions(ds9region, allow_empty)
+
+    def subcube_from_crtf(self, region, allow_empty=False):
+        """
+        Extract a masked subcube from a CRTF region
+        (only functions on celestial dimensions)
```
keflavich: I hope this line is no longer true when you're done

spectral_cube/spectral_cube.py
```diff
@@ -1912,6 +1912,122 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
...
+        ----------
+        region_list: `regions.Region` list
+            The region to extract
+        allow_empty: bool
+            If this is False, an exception will be raised if the region
+            contains no overlap with the cube
+        """
+        import regions
+
+        # Every region should be a `regions.PixelRegion` object.
```
keflavich: They aren't guaranteed to be yet.  Maybe rephrase the comment: "Convert all regions to `PixelRegion`s"

spectral_cube/spectral_cube.py
```diff
@@ -1912,6 +1912,122 @@ def subcube_from_ds9region(self, ds9region, allow_empty=False):
...
+                raise TypeError("'{}' should be `regions.Region` object".format(x))
+
+        # List of regions are converted to a `regions.CompoundPixelRegion` object.
+        compound_region = _compound_region(regs)
+
+        # Compound mask of all the regions.
+        mask = compound_region.to_mask()
+
+        # Checking for frequency/velocity range of each region.
+        ranges = [x.meta.get('range', None) for x in regs]
```
keflavich: ooh, this is a bit ugly, since they could technically be different....  we will need additional machinery to create the masks for each PPV sub-region instead of using the compound region (or at least, instead of *just* using the compound region)

sushobhana: Oh! Can you be a little more elaborative here? Like what technical differences can we face here and what machinery should we use to create the masks? Currently it just replicates the old `subcube_from_ds9region`.

spectral_cube/spectral_cube.py
sushobhana: that is why the following code converts `SkyRegion` objects to `PixelRegion` objects.
your rephrase is clearer though.

spectral_cube/spectral_cube.py
sushobhana: I am not so sure, since the wcs we use for conversion to pixel region is `cube.wcs.celestial` ?

spectral_cube/spectral_cube.py
keflavich: Right, but the old `subcube_from_ds9region` only did spatial.  With CRTF, we can do spatial-spatial-spectral.  So this is fine for the ds9 regions, but we want the CRTFs to allow us to do more.

spectral_cube/spectral_cube.py
keflavich: This is what I'm talking about on the other comment: CRTFs have spectral information too, not just spatial.  For the ds9 regions, they will only function on celestial dimensions, but CRTFs will function on celestial + spectral

sushobhana: oh yes ! I misread `celestial dimension` to `celestial coordinates`

spectral_cube/spectral_cube.py
sushobhana: Do the spectral dimensions have any system or convention ? I just try to find out the spectral slab range from the `range` key. What can we do more? You were also talking about technical differences?

spectral_cube/spectral_cube.py
sushobhana: I get you! sorry I was thinking something else.

spectral_cube/spectral_cube.py
sushobhana: Also, we can make use of the stokes parameters in CRTF regions.

keflavich: Yes, there is some important frame information in the `range`.  The `velocity_convention` is something that should be specified in the CRTF region... I'm not sure what the default is.  

So, for example, if you have a region with `range=[10 km /s, 20 km / s]` but the cube is in frequency (GHz, say), you would need to do `range.to(u.GHz, u.doppler_radio(rest_frequency)` if the velocity convention is radio...

sushobhana: `veltype` is for the velocity convention. The dafault value is the image value. 
`restfreq` is for rest frequency. The default says image value .
I will work on this.

sushobhana: https://help.almascience.org/index.php?/Knowledgebase/Article/View/86/0/what-are-the-frequency-reference-frames-in-casa I got something here.

sushobhana: Now, I get why we need a machinery here. Let's just see if astropy has handle all this.

spectral_cube/spectral_cube.py
```diff
@@ -2028,6 +2033,77 @@ def subcube_from_regions(self, region_list, allow_empty=False):
...
+    def _velocity_freq_conversion(self, ranges, veltypes, restfreqs):
+
+        header = self.wcs.to_header()
+
+        # Obtaining rest frequency of the cube in GHz.
+        restfreq_cube = header['restfrq'] * u.Hz
+
+        CTYPE3 = header['CTYPE3']
+
+        if 'RAD' in CTYPE3:
```
keflavich: instead of these if statements, use this:
https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_axis.py#L73

spectral_cube/spectral_cube.py
```diff
@@ -2028,6 +2033,77 @@ def subcube_from_regions(self, region_list, allow_empty=False):
...
+                     )]
+
+        def doppler_gamma(restfreq):
+            return [(u.GHz, u.km / u.s,
+                        lambda x: si.c.to_value("km/s") * (1 + (x / restfreq)**2) /
+                                  (2 * x / restfreq),
+                        lambda x: restfreq * ((x +
+                                  np.sqrt(x ** 2 - si.c.to_value("km/s") ** 2))
+                                  / si.c.to_value("km/s"))
+                     )]
```
keflavich: These are nicely done, but they are all implemented in astropy.  You can use this code:
https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/spectral_axis.py#L90
to determine which to use

spectral_cube/spectral_cube.py
sushobhana: Thanks!

spectral_cube/spectral_cube.py
sushobhana: I checked the code, the velocity convention `RADIO`, `RELATIVISTIC`, `OPTICAL` are handled there, but not 
these : `Z`, `BETA`, `GAMMA`

keflavich: What are z, beta, gamma?  I've never heard of these before.  Maybe they should be implemented in astropy...

keflavich: Right, `z` is redshift, `gamma` is the relativistic gamma, beta... I don't know.  These should be implemented in astropy, but in the meantime, maybe move all of the spectral-handling code to `spectral_axis`.

spectral_cube/spectral_axis.py
```diff
@@ -153,6 +153,34 @@ def get_rest_value_from_wcs(mywcs):
...
+
+def doppler_z(restfreq):
+    return [(u.GHz, u.km / u.s,
+                lambda x: (restfreq - x) / x,
+                lambda x: restfreq / (1 + x))]
+
+
+def doppler_beta(restfreq):
+    return [(u.GHz, u.km / u.s,
+            lambda x: si.c.to_value('km/s') * (1 - (x / restfreq) ** 2)
```
migueldvb: `si.c` should be `constants.si.c` ?

spectral_cube/spectral_axis.py
```diff
@@ -153,6 +153,34 @@ def get_rest_value_from_wcs(mywcs):
...
+                    / (1 + (x / restfreq) ** 2),
+            lambda x: np.sqrt(
+                (restfreq * ((si.c.to_value("km/s") - x) / (x + si.c.to_value("km/s")))))
```
migueldvb: `si.c` should be `constants.si.c` ?

spectral_cube/spectral_axis.py
```diff
@@ -153,6 +153,34 @@ def get_rest_value_from_wcs(mywcs):
...
+            )]
+
+
+def doppler_gamma(restfreq):
+    return [(u.GHz, u.km / u.s,
+                lambda x: si.c.to_value("km/s") * (1 + (x / restfreq)**2) /
+                            (2 * x / restfreq),
+                lambda x: restfreq * ((x / si.c.to_value("km/s") +
+                            np.sqrt((x / si.c.to_value("km/s")) ** 2 - 1)))
```
migueldvb: numpy is not imported in spectral_axis.py

sushobhana: sorry, i must have missed this while shifting the code here. No, worries I have added tests now.

keflavich:
@sushobhana could you fix the minor conflicts?

Is this ready for final review and/or merge?

spectral_cube/spectral_cube.py
sushobhana: If we want this method in regions, we better get this merged to regions/master before we merge this?

sushobhana: I think has this method has a very specific purpose (to convert a region list to a single `Region` object),
so do you think there is a good need to move to regions (having second thoughts here.)

sushobhana:
@keflavich Ready for final review!

spectral_cube/spectral_cube.py
keflavich: You're right, the function is more about manipulating a list into a region.  Change the function name to `_regionlist_to_single_region` and leave it here.

sushobhana:
May be I will change the docs here or a separate PR?

keflavich:
How about a separate PR.  It makes reviewing a little easier

sushobhana:
OK!

sushobhana:
@keflavich Ready to merge?

PR #489, opened 2018-06-16
Minor changes in docs

migueldvb:


keflavich:
Neat, thanks for this fix!  I suppose we didn't know how to do this concisely when we wrote the docs.

keflavich:
There are two failures in python2, which I suspect may be because python2 is tied to an older astropy version that doesn't handle WCS as well.... but I'm not really sure.

keflavich:
If https://github.com/radio-astro-tools/spectral-cube/pull/490 passes, I recommend merging that into this one; I hope it will resolve the problem.  My local tests on py2.7 w/np1.13 passed using your modifications.

migueldvb:
I have merged #490 and got one error. It seemed to be a change in the indentation of the output from the `print(cube)` command in http://spectral-cube.readthedocs.io/en/latest/index.html#quick-start.  I hope this build job will pass now.

PR #494, opened 2018-07-25
Add spectral-axis sigma clipping

keflavich:
This is a tool needed in statcont, and it can use the parallelization we have built in spectral cube.

This PR also smuggles in:
 
 * exposure of the parallel-apply-along-axis method
 * a new unitless_filled_data view

### DIFF
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -280,6 +280,19 @@ class MaskableArrayMixinClass(object):
                               self.unit, copy=False)
         return self.filled_data[:]
 
+    @cube_utils.slice_syntax
+    def unitless_filled_data(self, view):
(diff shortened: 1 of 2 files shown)


spectral_cube/base_class.py
```diff
@@ -280,6 +280,19 @@ def filled(self, fill_value=None):
                               self.unit, copy=False)
         return self.filled_data[:]
 
+    @cube_utils.slice_syntax
+    def unitless_filled_data(self, view):
+        """
+        Return a portion of the data array, with excluded mask values
+        replaced by :meth:`fill_value`.
```
keflavich: Is `fill_value` really a method...?  Maybe it is right to reference it this way?

Issue #492, opened 2018-07-16
Require >v0.2 of regions


e-koch:
@sushobhana, @keflavich - With the changes in #488, I found that running the tests with v0.2 of regions causes several failures in `test_subcubes.py`. Most of the failed tests fail on errors in the code, but this one actually returns an incorrect value:
```
____________________________________________ test_regions_spectral ____________________________________________

    @pytest.mark.skipif('not scipyOK', reason='Could not import scipy')
    @pytest.mark.skipif('not regionsOK', reason='Could not import regions')
    def test_regions_spectral():
        cube, data = cube_and_raw('adv.fits')
        rf_cube = get_rest_value_from_wcs(cube.wcs).to("GHz",
                                                             equivalencies=u.spectral())

        # content of image.reg
        regpix = regions.RectanglePixelRegion(regions.PixCoord(0.5, 1), width=4, height=2)

        # Velocity range in doppler_optical same as that of the cube.
        vel_range_optical = u.Quantity([-318 * u.km/u.s, -320 * u.km/u.s])
        regpix.meta['range'] = list(vel_range_optical)
        sc1 = cube.subcube_from_regions([regpix])
        scsum1 = sc1.sum()

        freq_range = vel_range_optical.to("GHz",
                                          equivalencies=u.doppler_optical(rf_cube))
        regpix.meta['range'] = list(freq_range)
        sc2 = cube.subcube_from_regions([regpix])
        scsum2 = sc2.sum()

        regpix.meta['restfreq'] = rf_cube
        vel_range_gamma = freq_range.to("km/s", equivalencies=doppler_gamma(rf_cube))
        regpix.meta['range'] = list(vel_range_gamma)
        regpix.meta['veltype'] = 'GAMMA'
        sc3 = cube.subcube_from_regions([regpix])
        scsum3 = sc3.sum()

        vel_range_beta = freq_range.to("km/s",
                                        equivalencies=doppler_beta(rf_cube))
        regpix.meta['range'] = list(vel_range_beta)
        regpix.meta['veltype'] = 'BETA'
        sc4 = cube.subcube_from_regions([regpix])
        scsum4 = sc4.sum()

        vel_range_z = freq_range.to("km/s",
                                        equivalencies=doppler_z(rf_cube))
        regpix.meta['range'] = list(vel_range_z)
        regpix.meta['veltype'] = 'Z'
        sc5 = cube.subcube_from_regions([regpix])
        scsum5 = sc5.sum()

[...]

PR #495, opened 2018-07-30
Doc on using regions.

sushobhana:
also addresses #492

### DIFF
--- a/docs/manipulating.rst
+++ b/docs/manipulating.rst
@@ -96,14 +96,14 @@ extract the cube in the spectral dimension as well.
 This extracts DS9 region(s) from ``file.reg`` file. The `~regions.read_ds9`
 parses the ds9 file and converts it to a list of `~regions.Region` object::
 
-    >>> import regions
+    >>> import regions # doctest: +SKIP
     >>> region_list = regions.read_ds9('file.reg')  # doctest: +SKIP
     >>> sub_cube = cube.subcube_from_regions(region_list)  # doctest: +SKIP
 
 This extracts CRTF regions from ``file.crtf`` file. The `~regions.read_crtf`
 also returns a list of `~regions.Region` object::
 
-    >>> import regions
+    >>> import regions # doctest: +SKIP
     >>> region_list = regions.read_crtf('file.reg')  # doctest: +SKIP
     >>> sub_cube = cube.subcube_from_regions(region_list)  # doctest: +SKIP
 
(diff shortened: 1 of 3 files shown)


docs/manipulating.rst
```diff
@@ -79,31 +79,65 @@ Numpy slicing notation::
...
+
+This extracts CRTF regions from ``file.crtf`` file. The `~regions.read_crtf`
+also returns a list of `~regions.Region` object::
+
+    >>> import regions # doctest: +SKIP
+    >>> region_list = regions.read_crtf('file.reg')  # doctest: +SKIP
+    >>> sub_cube = cube.subcube_from_regions(region_list)  # doctest: +SKIP
 
 If you want to loop over individual regions with a single region file, you need to convert the individual
 region to a shape list due to limitations in pyregion::
```
keflavich: Since we're no longer using `pyregion`, you should edit this to show that you *don't* need to create `pyregion.ShapeList` objects any more.

docs/manipulating.rst
```diff
@@ -79,31 +79,65 @@ Numpy slicing notation::
...
+    >>> region_str = "galactic; circle(0, 0, 0.1)"  # doctest: +SKIP
+    >>> sub_cube = cube.subcube_from_ds9region(region_str)  # doctest: +SKIP
+
+Similarly, you can also use a CRTF region string::
+
+    >>> region_str = "circle[[0deg, 0deg], 0.1deg], coord=galactic, range=[150km/s, 300km/s]"  # doctest: +SKIP
+    >>> sub_cube = cube.subcube_from_crtfregion(region_str)  # doctest: +SKIP
+
+Not only a `regions.Region` list extracts the cube in the spatial dimension
+but also helps cut through the spectral dimension . The ``meta`` attribute
```
keflavich: This sentence needs to be rephrased, and there is a space before the final `.`

docs/spectral_extraction.rst
```diff
@@ -40,22 +40,35 @@ mask from scratch and apply it to the data.::
...
+`~spectral_cube.spectral_cube.SpectralCube.subcube_from_crtfregion`::
+
+    >>> crtf_str = 'circle[[19:23:43.907, +14:30:34.66], 3"], coord=fk5, range=[150km/s, 300km/s]'  # doctest: +SKIP
+    >>> subcube = cube.subcube_from_crtfregion(crtf_str)  # doctest: +SKIP
+    >>> spectrum = subcube.mean(axis=(1, 2))  # doctest: +SKIP
+
+We can also use a list of `~regions.Region` object to extract a subcube using
+`~spectral_cube.spectral_cube.SpectralCube.subcube_from_regions`.
+You can convert the ds9 string into `~regions.Region` object using
+`~regions.DS9Parser` or directly read from a file using `~regions.read_ds9`::
```
keflavich: This example doesn't use a string, right?  So maybe just leave it out.

Also, the example as you have it now only uses a single region.  We want to show the use case involving a list of regions, e.g., a circle and a square with partial overlap.

PR #496, opened 2018-08-13
CASA: Fix import

keflavich:
Apparently CASA's namespace has changed since we wrote their IO tools.


keflavich:
I'm looking forward to the day we can have CASA CI tests.

PR #500, opened 2018-10-01
Fix a bug with HDU writing

keflavich:
Projection & Slice HDUs were trying to write HDULists incorrectly; only Spectra should have those in general


keflavich:
This should have just been added to #499 since it fixes a bug introduced in there...

PR #501, opened 2018-11-19
Add documentation on how to use common_beam

keflavich:
We have all the stuff implemented, but there wasn't an example.

### DIFF
--- a/docs/smoothing.rst
+++ b/docs/smoothing.rst
@@ -27,6 +27,16 @@ single-resolution :class:`~spectral_cube.SpectralCube` instances, but for a
 :class:`~spectral_cube.VaryingResolutionSpectralCube`, the convolution kernel
 will be different for each slice.
 
+Common Beam selection
+---------------------
+You may want to convolve your cube to the smallest beam that is still larger
+than all contained beams.  To do this, you can use the
+:func:`radio_beam.common_beam.commonbeam` tool.  For example::
+
+    common_beam = cube.beams.common_beam()
+    new_cube = cube.convolve_to(common_beam)
+
+
 Spectral Smoothing
 ==================
 
--- a/docs/smoothing.rst
+++ b/docs/smoothing.rst
@@ -28,10 +28,10 @@ single-resolution :class:`~spectral_cube.SpectralCube` instances, but for a
 will be different for each slice.
 
 Common Beam selection
----------------------
+^^^^^^^^^^^^^^^^^^^^^
 You may want to convolve your cube to the smallest beam that is still larger
 than all contained beams.  To do this, you can use the
-:func:`radio_beam.common_beam.commonbeam` tool.  For example::
+:func:`~radio_beam.common_beam.commonbeam` tool.  For example::
 
     common_beam = cube.beams.common_beam()
     new_cube = cube.convolve_to(common_beam)
@@ -39,7 +39,7 @@ than all contained beams.  To do this, you can use the
 Sometimes, you'll encounter the error "Could not find common beam to deconvolve
 all beams." This is a real issue, as the algorithms we have in hand so far do
 not always converge on a common containing beam.  You can try to change the
-tolerance used in the `~radio_beam.commonbeam.getMinVolEllipse` code by
+tolerance used in the :func:`~radio_beam.commonbeam.getMinVolEllipse` code by
 specifying ``tolerance=1e-5`` or a smaller number.  Doing so will slow down the
 code, but may result in convergence.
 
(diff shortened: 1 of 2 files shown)


docs/smoothing.rst
```diff
@@ -27,6 +27,22 @@ single-resolution :class:`~spectral_cube.SpectralCube` instances, but for a
...
+than all contained beams.  To do this, you can use the
+:func:`~radio_beam.common_beam.commonbeam` tool.  For example::
+
+    common_beam = cube.beams.common_beam()
+    new_cube = cube.convolve_to(common_beam)
+
+Sometimes, you'll encounter the error "Could not find common beam to deconvolve
+all beams." This is a real issue, as the algorithms we have in hand so far do
+not always converge on a common containing beam.  You can try to change the
+tolerance used in the :func:`~radio_beam.commonbeam.getMinVolEllipse` code by
```
e-koch: The "smaller" part might be confusing. We should recommend increasing the tolerance and/or epsilon to address this error. How about:

You can try to change the tolerance and epsilon used in the :func:`~radio_beam.commonbeam.getMinVolEllipse` code to get the algorithm to converge. Specifying `tolerance` will set the convergence criteria for the algorithm; smaller values will take longer to reach convergence. Try increasing the tolerance from `tolerance=1e-4` to get the algorithm to converge. `epsilon` sets the fraction to overestimate the beam perimeters. Increasing epsilon from the default `epsilon=5e-4` will ensure the algorithm converges at the expense of overestimating the common beam area. The smallest `epsilon` where convergence is reached should be used.

keflavich:
Uh... my experience is the opposite of what you suggested.  I had to _decrease_ the tolerance to get the algorithm to converge; I think tolerance is the "quitting criterion" that it iterates to, right?  I have experienced several cases where a tolerance of 1e-4 resulted in a failure, but a tolerance of 1e-5 took longer and succeeded.  I haven't played with epsilon as much.

I think we need more documentation about these criteria.

docs/smoothing.rst
keflavich: I disagree; see below.

But, you can also directly suggest changes now!
image

e-koch:
Interesting... It must be when the solution jumps slightly inside the valid solution region, giving a beam slightly too small that fails, but still satisfies the tolerance. So I guess tolerance could go both ways, which maybe isn't a helpful recommendation.

Marginally increasing epsilon is usually the what I've done to reach convergence. As long as it remains quite small (~<1e-3), the common beam area will be increased by less than one-pixel area.

keflavich:
Can you incorporate some of that discussion into the documentation, then?  It might be best to directly say "we don't know exactly what will result in a converged beam in all cases, try combinations of parameters..."

e-koch:
Added a full explanation for common beams in radio-beam: https://github.com/radio-astro-tools/radio-beam/pull/69. Is it worth having a brief explanation here, or just a link to the radio-beam docs?

keflavich:
A brief explanation here is warranted, since this has already proved to be a commonly-encountered error.

e-koch:
@keflavich -- I couldn't figure out how to make multi-line suggestions, so I just added a new commit

keflavich:
lgtm.  Ready to merge this, then?

e-koch:
:+1:

e-koch:
Links to radio-beam functions are still failing in the docs. I'm unsure if this is an issue here or whether the radio-beam docs aren't building properly.

docs/smoothing.rst
```diff
@@ -27,6 +27,51 @@ single-resolution :class:`~spectral_cube.SpectralCube` instances, but for a
 :class:`~spectral_cube.VaryingResolutionSpectralCube`, the convolution kernel
 will be different for each slice.
 
+Common Beam selection
+^^^^^^^^^^^^^^^^^^^^^
+You may want to convolve your cube to the smallest beam that is still larger
+than all contained beams.  To do this, you can use the
+`~radio_beam.Beams.common_beam` tool.  For example::
```
keflavich:
```suggestion
`~radio_beam.multiple_beams.Beams.common_beam` tool.  For example::
```

docs/smoothing.rst
```diff
@@ -27,6 +27,51 @@ single-resolution :class:`~spectral_cube.SpectralCube` instances, but for a
...
+    common_beam = cube.beams.common_beam()
+    new_cube = cube.convolve_to(common_beam)
+
+Sometimes, you'll encounter the error "Could not find common beam to deconvolve
+all beams." This is a real issue, as the algorithms we have in hand so far do
+not always converge on a common containing beam. There are two ways to get the
+algorithm to converge to a valid common beam:
+
+1. **Changing the tolerance.** - You can try to change the tolerance used in the
+`~radio_beam.commonbeam.getMinVolEllipse` code by
```
keflavich: this one should just work.

PR #509, opened 2018-12-03
Use memory mapping when filling data by default

keflavich:
TODO: Write some memory tracking tests!!!


keflavich:
@e-koch it's worth having another look at this one; using memmap for every slice resulted in a huge slowdown, so I think that's a bad idea, but the memory testing is a real thing.

I'm considering having `use_memmap` set to `False` by default for `_get_filled_data`, rather than trying to "cleverly" select based on cube size, since the latter is long-term a recipe for disaster.

keflavich:
I'm finally looking at this again... and the failures appear to be that there are some version combinations where memory isn't allocated.

keflavich:
Maybe you're right that the memory tracker is looking in the wrong place

keflavich:
Will merge once travis passes; last change was to make sure we don't downgrade astropy-helpers accidentally.

PR #518, opened 2019-01-10
Add `use_memmap` argument for `reproject` and boost minimum required reproject version

keflavich:
In December 2018, we implemented a not-in-memory version of reproject for cubes.  This PR allows spectral-cube to use that version.


keflavich:
coverage has me confused - the new lines should be covered...

PR #521, opened 2019-01-16
Keep Beam in LDOs

e-koch:
I encountered arithmetic cases where `OneDSpectrum`s were not keeping the beam information. The beam needed to be kept in `__array_finalize__`.

I added the same fix for the `VaryingResolutionOneDSpectrum` (VRODS), but the fix is a bit messy (@keflavich, see comments). Also added a check for whether the beams table equals the data shape.

Also added a bunch of tests for whether LDO classes are keeping the beams correctly. There are separate tests for VRODSs.

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -643,13 +643,15 @@ class Slice(Projection):
     pass
 
 
-class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
-                   SpectralAxisMixinClass, BeamMixinClass):
+class BaseOneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
+                       SpectralAxisMixinClass):
+    """
+    Properties shared between OneDSpectrum and VaryingResolutionOneDSpectrum.
+    """
 
     def __new__(cls, value, unit=None, dtype=None, copy=True, wcs=None,
                 meta=None, mask=None, header=None, spectral_unit=None,
-                fill_value=np.nan, wcs_tolerance=0.0,
-                beam=None, read_beam=False):
+                fill_value=np.nan, wcs_tolerance=0.0):
 
         #log.debug("Creating a OneDSpectrum with __new__")
 
@@ -680,20 +682,6 @@ class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
                 self._spectral_unit = u.Unit(self._header['CUNIT1'])
             elif self._wcs is not None:
                 self._spectral_unit = u.Unit(self._wcs.wcs.cunit[0])
-
-        if beam is None:
-            if "beam" in self.meta:
-                beam = self.meta['beam']
-            elif read_beam:
-                beam = cube_utils.try_load_beam(header)
-                if beam is None:
-                    warnings.warn("Cannot load beam from header.",
-                                  BeamWarning
-                                 )
-
-        if beam is not None:
-            self.beam = beam
-            self.meta['beam'] = beam
 
         return self
 
@@ -752,7 +740,7 @@ class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
 
     @property
     def header(self):
-        header = super(OneDSpectrum, self).header
+        header = super(BaseOneDSpectrum, self).header
 
         # Preserve the spectrum's spectral units
         if 'CUNIT1' in header and self._spectral_unit != u.Unit(header['CUNIT1']):
@@ -762,7 +750,6 @@ class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
             header['CUNIT1'] = self.spectral_axis.unit.to_string(format='FITS')
 
         return header
-
 
     @property
     def spectral_axis(self):
@@ -802,7 +789,6 @@ class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
         if filename is not None:
             pyplot.gcf().savefig(filename)
 
-
     def with_spectral_unit(self, unit, velocity_convention=None,
                            rest_value=None):
 
@@ -829,12 +815,15 @@ class OneDSpectrum(LowerDimensionalObject, MaskableArrayMixinClass,
                                        header=newheader)
 
     def __getitem__(self, key, **kwargs):
+        # Ideally, this could just be in VaryingResolutionOneDSpectrum,
+        # but it's about the code is about the same length by just
+        # keeping it here.
         try:
             kwargs['beams'] = self.beams[key]
-        except (AttributeError,TypeError):
+        except (AttributeError, TypeError):
             pass
 
-        new_qty = super(OneDSpectrum, self).__getitem__(key)
+        new_qty = super(BaseOneDSpectrum, self).__getitem__(key)
 
         if isinstance(key, slice):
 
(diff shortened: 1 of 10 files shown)


spectral_cube/lower_dimensional_structures.py
```diff
@@ -1058,23 +1082,26 @@ def _new_spectrum_with(self, data=None, wcs=None, mask=None, meta=None,
         fill_value = self._fill_value if fill_value is None else fill_value
         spectral_unit = self._spectral_unit if spectral_unit is None else u.Unit(spectral_unit)
 
-        if beams is None:
-            if hasattr(self, 'beams'):
-                kwargs['beams'] = beams
-        elif beam is None:
-            if hasattr(self, 'beam'):
+        if beams is None and not hasattr(self, 'beams'):
```
e-koch: @keflavich can you see a cleaner way to do this? Basically, I want to avoid checking for `self.beam` when it's a VRODS.

spectral_cube/lower_dimensional_structures.py
```diff
@@ -144,6 +144,11 @@ def __array_finalize__(self, obj):
         self._fill_value = getattr(obj, '_fill_value', np.nan)
         self._wcs_tolerance = getattr(obj, '_wcs_tolerance', 0.0)
 
+        if isinstance(obj, VaryingResolutionOneDSpectrum):
+            self._beams = getattr(obj, '_beams', None)
+        else:
+            self._beam = getattr(obj, '_beam', None)
```
keflavich: it might be better to have `VaryingResolutionOneDSpectrum` have an `__array_finalize__` that deletes the `_beam` object and adds the `_beams` object?  Or, have each subclass implement their own `__array_finalize__` that adds the appropriate type (after using super() to call this `finalize`)

spectral_cube/lower_dimensional_structures.py
```diff
@@ -1093,6 +1120,51 @@ def __getattribute__(self, attrname):
...
+        beams : `~radio_beam.Beams`
+            A new beams object.
+        '''
+
+        meta = self.meta.copy()
+        meta['beams'] = beams
+
+        return self._new_spectrum_with(beams=beams, meta=meta)
+
+    def with_beam(self, beam):
```
e-koch: I wanted to block the case of setting a single beam to a VRODS so there was no way to mistakenly go from VRODS to ODS. Raising the error could be an issue though... same with the re-defined `beam` parameter below.

spectral_cube/lower_dimensional_structures.py
```diff
@@ -1018,6 +1027,21 @@ def with_fill_value(self, fill_value):
         """
         return self._new_spectrum_with(fill_value=fill_value)
 
+    def with_beam(self, beam):
```
keflavich: should this method be defined on some even lower-level object...?  maybe not, just raising the possibility

spectral_cube/lower_dimensional_structures.py
```diff
@@ -1093,6 +1120,51 @@ def __getattribute__(self, attrname):
 
 class VaryingResolutionOneDSpectrum(OneDSpectrum, MultiBeamMixinClass):
     def __new__(cls, value, beams=None, **kwargs):
+        out = super(VaryingResolutionOneDSpectrum, cls).__new__(cls, value, **kwargs)
         if beams is not None:
-            cls.beams = beams
-        return super(VaryingResolutionOneDSpectrum, cls).__new__(cls, value, **kwargs)
+            out.beams = beams
```
e-koch: Needed this to trigger the `beams.setter` checks below. Setting `cls.beams`, then running `super` didn't do this.

spectral_cube/lower_dimensional_structures.py
```diff
@@ -1093,6 +1120,51 @@ def __getattribute__(self, attrname):
...
+        meta['beams'] = beams
+
+        return self._new_spectrum_with(beams=beams, meta=meta)
+
+    def with_beam(self, beam):
+        '''
+        Re-define here so a single Beam cannot be passed.
+        '''
+        raise ValueError("For VaryingResolutionOneDSpectrum's, use with_beams "
+                         "instead of with_beam.")
```
keflavich: Right... there's not a way to "undefine" a method.  This implies that we should have a `BaseOneDSpectrum` object and have `OneDSpectrum` and `VaryingResolutionOneDSpectrum` inherit from it, like we do with Cubes.

keflavich:
@e-koch sorry we crossed chats.  I think my recommendations might be the right approach, though?

spectral_cube/lower_dimensional_structures.py
e-koch: Yeah. So we'd have `LDO` -> `BaseOneDSpectrum` -> `OneDSpectrum`, `VRODS`. And `beam` would have to be defined in `Projection`, too. Despite the multiple levels of inheritance, it will probably still be easier to understand...

spectral_cube/lower_dimensional_structures.py
e-koch: This would be solved by inheriting from a `BaseOneDSpectrum`, too

e-koch:
Some tests are hanging/failing before starting. We probably need to make some updates with the latest changes in ci-helpers.

keflavich:
Yes, see #517?

e-koch:
Tests are mostly running now. One memory test is failing with np 1.11 (https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/480522946#L1583). Maybe another version problem?

e-koch:
@keflavich -- Can you take another look at this PR? I've added a new `BeamMixinClass` for single beams and a `BaseOneDSpectrum`

spectral_cube/lower_dimensional_structures.py
```diff
@@ -1058,41 +1058,136 @@ def _new_spectrum_with(self, data=None, wcs=None, mask=None, meta=None,
...
+            hdu = self.hdu
+
+        beamhdu = cube_utils.beams_to_bintable(self.beams)
+
+        return HDUList([hdu, beamhdu])
+
+    def _new_spectrum_with(self, data=None, wcs=None, mask=None, meta=None,
+                           fill_value=None, spectral_unit=None, unit=None,
+                           header=None, wcs_tolerance=None, beams=None,
+                           **kwargs):
```
keflavich: isn't most of this redundant?  Could we use `super()` here instead of rewriting all this code?  So we could have `BaseOneDSpectrum._new_spectrum_with()` that would take care of everything except the beams, and have the `beam` or `beams` handling be in the subclasses?

spectral_cube/tests/test_projection.py
```diff
@@ -141,6 +142,87 @@ def test_self_arith(LDO, data):
...
+
+
+def test_VRODS_with_beams():
+
+    exp_beams = Beams(np.arange(1, twelve_qty_1d.size + 1) * u.arcsec)
+
+    p = VaryingResolutionOneDSpectrum(twelve_qty_1d, copy=False, beams=exp_beams)
+    assert (p.beams == exp_beams).all()
+
+    p = p.with_beams(exp_beams * 2)
```
keflavich: is it a mistake that we're allowing beams to be multiplied?  I don't think that operation makes sense, and if you try to multiply a single beam by a value, it fails:
```
In [11]: bm*5
  File "<ipython-input-11-9c711351f2a3>", line 1, in <module>
    bm*5
  File "/Users/adam/repos/radio_beam/build/lib.macosx-10.7-x86_64-3.6/radio_beam/beam.py", line 345, in __mul__
    return self.convolve(other)
  File "/Users/adam/repos/radio_beam/build/lib.macosx-10.7-x86_64-3.6/radio_beam/beam.py", line 338, in convolve
    new_major, new_minor, new_pa = convolve(self, other)
  File "/Users/adam/repos/radio_beam/build/lib.macosx-10.7-x86_64-3.6/radio_beam/utils.py", line 126, in convolve
    (other.major * np.cos(other.pa))**2 +
AttributeError: 'int' object has no attribute 'major'
```

e-koch: Agreed. I'll just generate a new set of beams here.

spectral_cube/lower_dimensional_structures.py
e-koch: Fixed.

e-koch:
@keflavich - I think this is ready to merge.

PR #525, opened 2019-01-25
BUG: WCS downsampling had incorrect CRPIX

keflavich:
Serious bug: this is causing a shift in downsampled data


keflavich:
@astrofrog this is a fairly serious bug in the WCS downsampling; do you recall if we've used this code elsewhere?

Issue #533, opened 2019-02-21
`update_function` removed from `spatial_smooth` method


brechmos-stsci:
We are using the `update_function` parameter of `def spatial_smooth()` to get updates on the smoothing as it progresses.  It looks like the refactor of the smoothing in 0.4.4 removed this ability, but not sure if that was a decision or an oops. 

The traceback is:
```
---------------------------------------------------------------------------                                                                      
TypeError                                 Traceback (most recent call last)                                                                      
<ipython-input-21-db517e558904> in <module>
----> 1 cube.spatial_smooth(k, update_function=lambda x: print('here'))

/opt/miniconda3/envs/cubeviz-dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in spatial_smooth(self, kernel, convolve, **kwargs)  
   2604
   2605         newcube = self.apply_function_parallel_spatial(_gsmooth_image,                                                                   
-> 2606                                                        **kwargs)
   2607
   2608         return newcube

/opt/miniconda3/envs/cubeviz-dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in apply_function_parallel_spatial(self, function, num
_cores, verbose, use_memmap, parallel, **kwargs)
   2789                                                   num_cores=num_cores,                                                                   
   2790                                                   use_memmap=use_memmap,                                                                 
-> 2791                                                   **kwargs)
   2792
   2793     def apply_function_parallel_spectral(self,

/opt/miniconda3/envs/cubeviz-dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in _apply_function_parallel_base(self, iteration_data,
 function, applicator, num_cores, verbose, use_memmap, parallel, memmap_dir, **kwargs)                                                           
   2726
   2727             for arg in iteration_data:
-> 2728                 applicator(arg, outcube, function, **kwargs)
   2729                 pbu()
   2730

/opt/miniconda3/envs/cubeviz-dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in _apply_spatial_function(arguments, outcube, functio
n, **kwargs)
    118
    119     if np.any(includemask):
--> 120         outcube[ii, :, :] = function(img, **kwargs)
    121     else:
    122         outcube[ii, :, :] = img

/opt/miniconda3/envs/cubeviz-dev/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in _gsmooth_image(img, **kwargs)                     
[...]

PR #534, opened 2019-02-21
Fix for issue #533 - add back update_function

keflavich:
update_function was being used by other packages, and we deprecated it without warning them, which is a bug.

It turns out keeping `update_function` around is a bit of a pain, though, so let's see how this works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
         # overidden underneath. They are defined here in order to save having
         # to repeat them for all configurations.
         - PYTHON_VERSION=3.6
-        - SETUP_CMD='test'
+        - SETUP_CMD='test -vv'
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
         - CONDA_DEPENDENCIES='sip<4.19 aplpy glue-core scipy'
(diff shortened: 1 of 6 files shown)


keflavich:
@brechmos-stsci this *should* work, but `update_pb` must be picklable if you're using the parallel version.

keflavich:
@brechmos-stsci Any word on whether this solves your issue?

brechmos-stsci:
Sorry was away the past few days, will test today.

brechmos-stsci:
I did a `pip install git+git://github.com/keflavich/spectral-cube@update_function_callback` and then ran the Guassian and median smoothing. They both worked without crashing -- so, it looks good!
Thanks for the fix.

keflavich:
At least one of the test failures is apparently because of a problem in `joblib` being incompatible w/python 2.7.  However, afaict, it is supposed to be 2.7-compatible.  I'll probably choose to ignore that failure, but spectral-cube needs to continue to support 2.7 because of CASA for at least a year.

The other failures look like travis is just giving up?

keflavich:
The coverage decrease is a lie, it's because I can't run tests on travis =(

keflavich:
py27 now fails on glue import, so I'll delete that.  The rest fail at matplotlib import.

PR #535, opened 2019-02-22
CASA install: use subprocess instead of pip.main

keflavich:


keflavich:
Closing: we'll test w/casa6 instad

PR #537, opened 2019-02-27
fix for issue where num_cores = None (default?) raises an exception

keflavich:
Some previous PR... maybe even an unmerged one? broke the handling of `num_cores=None`


keflavich:
This was just a bug in #533, so I merged it there.

PR #544, opened 2019-03-06
Remove debug statements

keflavich:
I left some debug statements in #534 that I forgot about because I had to skip those tests on CI.


keflavich:
ffs... adding print statements _cannot_ cause these builds to fail.

PR #543, opened 2019-03-06
CASA images: read beam

keflavich:
When loading images (2D cubes) from CASA images, and maybe in other cases, the beam isn't loaded.

### DIFF
--- a/spectral_cube/io/casa_image.py
+++ b/spectral_cube/io/casa_image.py
@@ -197,7 +197,7 @@ def load_casa_image(filename, skipdata=False,
         cube.allow_huge_operations = True
 
     elif wcs.naxis == 4:
-        data, wcs = cube_utils._split_stokes(data.T, wcs)
+        data, wcs = cube_utils._split_stokes(data, wcs)
         mask = {}
         for component in data:
             data_, wcs_slice = cube_utils._orient(data[component], wcs)
--- a/spectral_cube/io/casa_image.py
+++ b/spectral_cube/io/casa_image.py
@@ -114,12 +114,11 @@ def load_casa_image(filename, skipdata=False,
     # read in the data
     if not skipdata:
         # CASA data are apparently transposed.
-        # (however, this is supposed to be handled in orient_data)
-        data = ia.getchunk().reshape(ia.shape())
+        data = ia.getchunk().transpose()
 
     # CASA stores validity of data as a mask
     if not skipvalid:
-        valid = ia.getchunk(getmask=True).reshape(ia.shape())
+        valid = ia.getchunk(getmask=True).transpose()
 
     # transpose is dealt with within the cube object
 
--- a/spectral_cube/io/casa_image.py
+++ b/spectral_cube/io/casa_image.py
@@ -114,7 +114,8 @@ def load_casa_image(filename, skipdata=False,
     # read in the data
     if not skipdata:
         # CASA data are apparently transposed.
-        data = ia.getchunk().reshape(ia.shape()).transpose()
+        # (however, this is supposed to be handled in orient_data)
+        data = ia.getchunk().reshape(ia.shape())
 
     # CASA stores validity of data as a mask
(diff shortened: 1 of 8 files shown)


keflavich:
todo: rebase/merge against #541

keflavich:
This is untested and will remain so until we can get access to casatools wheels.  That's not a great way to design things, but even running tests isn't straightforward for me...

that said, I'll try to make some tests before merging this, even if they won't run.

keflavich:
related: https://github.com/radio-astro-tools/radio-beam/pull/77

keflavich:
after a lot of back and forth, the tests pass on the machine I've been able to test on.

keflavich:
@e-koch could you have a quick look over this and https://github.com/radio-astro-tools/radio-beam/pull/77 and make sure I'm not doing anything blatantly stupid?

spectral_cube/io/casa_image.py
```diff
@@ -111,11 +113,12 @@ def load_casa_image(filename, skipdata=False,
 
     # read in the data
     if not skipdata:
-        data = ia.getchunk().reshape(ia.shape())
+        # CASA data are apparently transposed.
+        data = ia.getchunk().transpose()
```
e-koch: I'm wondering if a non-standard stokes axis screws this up somehow. Maybe we've already dealt with this for `StokesSpectralCube`. Just noting it here.

spectral_cube/tests/data/make_test_cubes.py
```diff
@@ -26,10 +26,10 @@ def transpose(d, h, axes):
     beams = np.recarray(4, dtype=[('BMAJ', '>f4'), ('BMIN', '>f4'),
                                   ('BPA', '>f4'), ('CHAN', '>i4'),
                                   ('POL', '>i4')])
-    beams['BMAJ'] = [0.1,0.2,0.3,0.4] # arcseconds
```
e-koch: Ah. Interesting. Good catch!

spectral_cube/tests/test_casafuncs.py
```diff
@@ -73,19 +117,45 @@ def test_casa_mask():
...
+    if os.path.exists(maskpath):
+        os.system('rm -rf {0}'.format(maskpath))
+
+    # in this case, casa.mask is the name of the mask, not its path
+    make_casa_mask(cube, path('casa.mask'), append_to_image=True,
+                   img=path('casa.image'), add_stokes=False,
+                   overwrite=True)
+
+    assert os.path.exists(path('casa.image/casa.mask'))
+
```
e-koch: Delete `casa.mask` and `casa.image` here. Unless they are needed in another test?

spectral_cube/tests/test_casafuncs.py
```diff
@@ -73,19 +117,45 @@ def test_casa_mask():
...
+    assert hasattr(cube, 'beam')
 
-    make_casa_mask(cube, 'casa.mask', append_to_image=True,
-                   img='casa.image', add_stokes=False)
+    cube_beams = SpectralCube.read(path('casa_adv_beams.image'), format='casa_image')
 
-    assert os.path.exists('casa.image/casa.mask')
+    assert hasattr(cube_beams, 'beams')
+    assert isinstance(cube_beams, VaryingResolutionSpectralCube)
```
e-koch: File clean-up here, too?

spectral_cube/io/casa_image.py
keflavich: Yeah, we had always had a `.transpose()` in stokes, but I think it turns out that CASA _always_ transposes the data, and we had never tested against stokesless cubes before (since CASA doesn't usually produce them...)

keflavich: (it was `.T`, not `.transpose()`, if you go hunting)

e-koch: Oh ok.

PR #551, opened 2019-04-19
Channel Maps

keflavich:
Replaces #195 as a tool for making channel maps in a grid

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -3275,6 +3275,68 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         return self._new_cube_with(data=dsarr, wcs=newwcs,
                                    mask=BooleanArrayMask(mask, wcs=newwcs))
 
+    def plot_channel_maps(self, output_file, nx, ny, channels, levels,
+                          figsize=(20,20), decimals=3, zoom=1):
+        """
+        Make channel maps from a spectral cube
+
+        Parameters
+        ----------
+        input_file : str
+            Name of the input spectral cube
+        output_file : str
+            Name of the matplotlib plot
+        nx, ny : int
+            Number of sub-plots in the x and y direction
+        channels : list
+            List of channels to show
+        levels : list
+            List of levels to show
+        figsize : tuple, optional
+            Figure size for matplotlib
+        decimals : int, optional
+            Number of decimal places to show in spectral value
+        zoom : int, optional
+            How much to zoom in. In future versions of this function, the
+            pointing center will be customizable.
+        """
+
+        import matplotlib.pyplot as plt
+
+        if len(channels) != nx * ny:
+            raise ValueError("Number of channels should be equal to nx * ny")
+
+        # Read in spectral cube and get spectral axis
+        spectral_axis = self.spectral_axis
+
+        fig = plt.figure(figsize=figsize)
+
+        sizey, sizex = self.shape[1:]
+        cenx = sizex / 2.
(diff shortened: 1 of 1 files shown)


keflavich:
Example:
```
cube = SpectralCube.read('/Users/adam/work/w51/alma/FITS/longbaseline/velo_cutouts/w51e2e_csv0_j2-1_r0.5_medsub.fits')
subcube = cube.spectral_slab(16*u.km/u.s, 87*u.km/u.s)

norm = vis.ImageNormalize(subcube,
                          interval=vis.ManualInterval(-0.01, 0.1),
                          stretch=vis.LogStretch())

fig = pl.figure(1, figsize=(18,18))
subcube.plot_channel_maps(5, 5, np.arange(25), norm=norm,
                          cmap='gray_r',
                          fig=fig,
                          contourkwargs=dict(
                          levels=[0.02,0.04,0.06,0.08,0.1,0.125,0.15],
                          colors=['w']*10,
                          linewidths=[0.5]*10,)
                         )
```

image

keflavich:
Open question: how do we force a tight layout (no whitespace horizontally or vertically between figures) without forcing a particular figure size?  This is definitely possible, but every stackoverflow question shows how to "work around" the issue by forcing a specific figure size, which I have defined here as not an acceptable solution, and the matplotlib gallery shows a bunch of examples that are incompatible (afaict) with `projection=`.

@astrofrog, any ideas?  You've probably dug into this as much as anyone.

keflavich:
Nico added some suggestions about tightening the layout, it now looks like this:
image


I've pseudo-automated the color selection for labels, but there ought to be a much better way to do this (e.g., https://gamedev.stackexchange.com/questions/38536/given-a-rgb-color-x-how-to-find-the-most-contrasting-color-y) that should be incorporated into matplotlib, I'd think.

Do we want to auto-add ra/dec labels of some sort?  We need to add colorbar.

e-koch:
:+1:  RA/Dec labels along the edges seem like a good default.

keflavich:
how should we decide on the labels?  Does astropy have a mapping from wcstype -> conventional label?  It should... I just don't know immediately

GiantMolecularCloud:
I'd suggest setting a default color for the velocity/frequency labels because the colormap also has a default. E.g. red labels will work on any grayscale map.
Imho: If you want to set another cmap yourself, it's your problem to find a matching overlay color.

GiantMolecularCloud:
regarding this: `fig.subplots_adjust(margin,margin,1.-margin,1.-margin,0.,0.)`
Interesting to see that it works without setting the figure margins explicitly. For my tests, I had to force equal margins around the grid. This might depend on the default margins in matplotlibrc.

GiantMolecularCloud:
tight_layout breaks the alignment of the panels and needs to be removed.
`savefig(... bbox_inches='tight')`still works though and crops some whitespace.

keflavich:
hm, both with and without `tighten=True` look the same for me.  Maybe this is backend-dependent?  That would suck.

```
axes = cube.plot_channel_maps(3,3,np.arange(0,80,9), output_file='test.pdf', tighten=False, savefig_kwargs={'bbox_inches':'tight'})
axes = cube.plot_channel_maps(3,3,np.arange(0,80,9), output_file='test_tightened.pdf', tighten=True, savefig_kwargs={'bbox_inches':'tight'})
```

GiantMolecularCloud:
Maybe. I've used Qt5Agg for my tests.

keflavich:
Ready to merge if we can get tests passing; see #558

PR #552, opened 2019-04-30
Fix world_extrema for 2D objects

e-koch:
@keflavich - This was broken while trying to mitigate the memory usage for finding extrema.

Do we have an axis type helper function somewhere? The current fix won't work for an actual polarization axis in a cube.

### DIFF
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -195,13 +195,23 @@ class SpatialCoordMixinClass(object):
     @property
     @cached
     def world_extrema(self):
-        corners = [(0, self.shape[2]-1),
-                   (self.shape[1]-1, 0),
-                   (self.shape[1]-1, self.shape[2]-1),
-                   (0,0)]
-        latlon_corners = [self.world[0, y, x] for y,x in corners]
-        lon = u.Quantity([x for z,y,x in latlon_corners])
-        lat = u.Quantity([y for z,y,x in latlon_corners])
+        if len(self.shape) == 2:
+            corners = [(0, self.shape[1]-1),
+                       (self.shape[0]-1, 0),
+                       (self.shape[0]-1, self.shape[1]-1),
+                       (0, 0)]
+            latlon_corners = [self.world[y, x] for y,x in corners]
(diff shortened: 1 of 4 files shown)


keflavich:
Good question.  I think such a helper would live in `astropy.wcs.utils`?

e-koch:
Yeah. I'm going to generalize finding the spatial axes before merging. Should be able to find the correct slice from `wcs.sub()` and use that.

spectral_cube/base_class.py
```diff
@@ -108,6 +113,20 @@ def _raise_wcs_no_celestial(self):
...
+        '''
+
+        self._raise_wcs_no_celestial()
+
+        wcs_cel_axis = [self.wcs.world_axis_physical_types.index(axtype)
+                        for axtype in
+                        self.wcs.celestial.world_axis_physical_types]
+
+        # Swap to numpy ordering
+        return [wcs2np[val] for val in wcs_cel_axis][::-1]
```
keflavich: Is this line doubling down on the axis reversal with both `wcs2np[val]` and `[::-1]`?

e-koch: `wcs_cel_axis` comes out backward. I moved the `[::-1]` onto `wcs_cel_axis` so it's more clear what is reversed.

keflavich: if it comes out backward... then isn't `wcs_cel_axis` in numpy order to begin with?  Theoretically, what you're doing is more readable, but afaict, you are reversing the order twice, which is just more confusing.

It might be most helpful to include an example in the comments showing what the expected types are, e.g.:
```
# for an RA, Dec WCS we expect [1, 0] <-> [Dec, RA]
```
or something along those lines.

e-koch: There's a 2 to 0 mapping for the spatial-x axis in cubes so it's not quite just reversed. But that means this only works for 3D cubes and isn't general... so it needs to be changed regardless

keflavich: Ah!  Right, ok!

PR #556, opened 2019-07-19
Small improvement to most commonly seen error message

keflavich:
Refers users to the docs and gives suggestions on how to do things w/big data.

cc @e-koch, @low-sky - very brief; could you review?  Other ideas?

### DIFF
--- a/spectral_cube/utils.py
+++ b/spectral_cube/utils.py
@@ -4,6 +4,8 @@ from functools import wraps
 
 from astropy.utils.exceptions import AstropyUserWarning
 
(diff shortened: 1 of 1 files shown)


keflavich:
Yeah, I was considering linking directly there, but I didn't like how long the URL was... it might wrap on most peoples' terminals.  Any ideas?

e-koch:
Ah, good point. I'm not sure how to deal with that. The shorter URL should be good, then.

keflavich:
Are there any good (like, not spammy) URL shorteners?  Or... maybe we just leave it as is and hope people can read the docs....

Issue #560, opened 2019-07-31
UserWarning: Could not parse beam information from header. For near-IR spectral cubes.


lib-j:
Hi, 

I am using spectral-cube to analyse a near-IR spectral cube from KMOS and keep getting an error message about not parsing beam information from header. This seems to assume  all spectral cubes are radio based.  Could a check be added to see if the data is in sub-mm/radio data before giving this warning?  Or a way for the user to flag no beam information is present.

anaconda3/envs/astro36/lib/python3.6/site-packages/spectral_cube/spectral_cube.py:246: UserWarning: Could not parse beam information from header.  Exception was: NoBeamException('No BMAJ found and does not appear to be a CASA/AIPS header.',)
  "  Exception was: {0}".format(ex.__repr__()))

keflavich:
@lib-j Thanks.  Yes, we can just remove this warning, it is not terribly helpful for anyone as is, I think.

In the meantime, you can suppress the warning by doing something like:
```
import warnings
warnings.filterwarnings('ignore', module=spectral_cube)
```
I think... I'll have to check the exact syntax

keflavich:
Looks like the appropriate syntax may (also?) be:
```
warnings.filterwarnings(action='ignore', category=spectral_cube.utils.BeamWarning)
```

lib-j:
Fantastic. Thank you :)

keflavich:
Let's leave this open until we actually remove the warning, though

e-koch:
We could remove this warning in favour of one for when `SpectralCube.beam` is `None`.

keflavich:
Yeah, I like that idea.  If you want to do it, please go ahead, otherwise I'll try to remember to try tomorrow....

(still trying to get CI to pass though =( )

keflavich:
closed by #561

PR #561, opened 2019-07-31
Fix warnings when a beam is not defined

e-koch:
Addresses #560. 

Changes the behaviour such that `SpectralCube.beam` raises a `NoBeamError` if a beam object is not defined. Removes the UserWarning in 560 that wasn't very helpful/informative.

### DIFF
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -745,7 +745,7 @@ class BeamMixinClass(object):
             raise NoBeamError("No beam is defined for this SpectralCube or the"
                               " beam information could not be parsed from the"
                               " header. A `~radio_beam.Beam` object can be"
-                              " defined using `cube.with_beam`.")
+                              " added using `cube.with_beam`.")
 
         return self._beam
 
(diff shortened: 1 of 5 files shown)


keflavich:
There are failures that are not doc failures but might be numpy version failures?  I'm uncomfortable with these precision-related errors; I don't know why they're appearing now.  However, they are below the precision of the methods we're testing, so I guess we push on?  I'm still going to wait a little..

e-koch:
That's worrying. I'll dig a little bit into that precision error.

e-koch:
Ok, I can reproduce this locally on the astropy dev branch ('4.0.dev25484').

e-koch:
@keflavich -- The unit values `cube.jtok_factors` was tested against had what I think was small rounding errors.

e-koch:
Also, `radio-beam`'s tests are fine w/ astropy v4dev

keflavich:
👍 could you rebase this now that #588 is done?

keflavich:
@e-koch could you rebase this again?  It looks like there are  40 commits but should only be ~8?

e-koch:
So the rebase said it went fine... but clearly it hasn't

e-koch:
Ok. Merging worked. _Should_ be ready if this passes.

PR #563, opened 2019-08-01
[Depends on upstream] Replace spectral_cube's SpectralAxisMixin w/specutils'

keflavich:
See https://github.com/astropy/specutils/pull/499

This will fail, but the failures should inform us about next steps.

My tests still hang locally....


e-koch:
(Hopefully) fixed hanging test with #564.

keflavich:
This is on hold until astropy implements APE14:
https://github.com/astropy/astropy-APEs/blob/master/APE14.rst

PR #568, opened 2019-08-09
Fix slicing to OneDSpectrum without a beam

e-koch:
I missed this in #561: When slicing a cube to `OneDSpectrum`, `self.beam` is called which raises a BeamError when it is not defined.


e-koch:
Tests are passing, except the astropy dev version is failing to install on travis.

Issue #279, opened 2016-02-13
Adding more functionality to Projection


e-koch:
`BaseNDClass` should allow for `Projection` and `Slice` objects to access to the spatial world coordinate abilities of `SpectralCube` (longitude, latitude, extrema, etc...).

keflavich:
Sounds kind of reasonable.  Want to make a checklist for things you think could be common to all?

e-koch:
WIth `world` generalized for 2D projections, all of these become trivial to include.

* [ ] spatial_coordinate_map
* [ ] world
* [ ] world_extrema
* [ ] flattened_world
* [ ] longitude_extrema
* [ ] latitude_extrema
* [ ] "subcube" like behaviour, where slicing with world coordinates is accepted

This probably is not an exhaustive list yet.

keflavich:
@astrofrog we're likely to be able to implement a lot of this, and so we probably will.  Again, I wonder to what degree this overlaps/clashes with `astropy.nddata`.  We're living heavily on the functionality-driven side in spectral-cube, but we've now abstracted enough that I expect there to be some competition and/or possibly reusable resources.

e-koch:
This will also require a `apply_numpy_function` type function for the lower dimensional objects. Using built-in numpy functions like `proj.mean()` returns a `Projection`, not a `Quantity`.

keflavich:
Perhaps instead that should mean that `BaseNDClass` would not include any of the `apply_numpy_function` etc. functionality, and that would only go into `SpectralCube`?  Since those are meant to provide means to deal with very large data sets, they aren't necessary for lower-dimensional objects.

keflavich:
How much of this has been done?  Some for sure...

e-koch:
Finished off in #575.

PR #575, opened 2019-09-05
Update flattened_world

e-koch:
@keflavich -- I was trying to finish off #279 by adding a `Projection.flattened_world` but the SC version of the function has its own problems.

`cube._mask._flattened` for a cube mask doesn't allow the same slicing as `cube.world`. `flattened_world` could just return the flattened version of whatever `world` returns for consistency.

### DIFF
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -216,21 +216,7 @@ class SpatialCoordMixinClass(object):
 
         self._raise_wcs_no_celestial()
 
-        if not self.wcs.is_celestial:
-            spec, lon, lat = self.world[view]
-            spec = self._mask._flattened(data=spec, wcs=self._wcs, view=view)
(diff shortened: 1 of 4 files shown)


spectral_cube/base_class.py
```diff
@@ -205,6 +205,19 @@ def world(self, view):
...
         return world[::-1]  # reverse WCS -> numpy order
 
+    def flattened_world(self, view=()):
+        """
+        Retrieve the world coordinates corresponding to the extracted flattened
+        version of the cube
+        """
+
+        # NOTE: this should be moved to SpatialCoordMixinClass once masks
+        # are implemented for lower dim objects - EK
```
keflavich: This is in `SpatialCoordMixinClass` now, no?

e-koch: Yup! will remove.

Issue #577, opened 2019-09-16
Spectral_smooth does not work


hongliliu:
Dear developer:
Recently, I tried smoothing my raw data cube in velocity axis with a Gaussian keneral using the function cube.spectral_smooth. However, I always received a error message. Below, I list my scripts and the associated error message after the code runs. Please help me out on how to solve this problem. Thank you very much in advance.
cube = sc.read('t13_reg1.fits')
fwhm_factor = np.sqrt(8*np.log(2))
current_resolution = 0.26569 * u.km/u.s
target_resolution = 0.5 * u.km/u.s
pixel_scale = 0.26569 * u.km/u.s
gaussian_width = ((target_resolution**2 - current_resolution**2)**0.5 /
                  pixel_scale / fwhm_factor)
kernel = Gaussian1DKernel(gaussian_width)
new_cube = cube.spectral_smooth(kernel)

---error message---
---> 10 new_cube = cube.spectral_smooth(kernel)

~/anaconda3/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in spectral_smooth(self, kernel, convolve, verbose, use_memmap, num_cores, **kwargs)
   2949                                                      use_memmap=use_memmap,
   2950                                                      verbose=verbose,
-> 2951                                                      **kwargs)
   2952 
   2953     def spectral_interpolate(self, spectral_grid,

~/anaconda3/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in apply_function_parallel_spectral(self, function, num_cores, verbose, use_memmap, parallel, **kwargs)
   2882                                                   verbose=verbose,
   2883                                                   num_cores=num_cores,
-> 2884                                                   **kwargs
   2885                                                  )
   2886 

~/anaconda3/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in _apply_function_parallel_base(self, iteration_data, function, applicator, num_cores, verbose, use_memmap, parallel, memmap_dir, update_function, **kwargs)
   2762 
   2763             for arg in iteration_data:
-> 2764                 applicator(arg, outcube, function, **kwargs)
   2765                 pbu()
   2766 

~/anaconda3/lib/python3.6/site-packages/spectral_cube/spectral_cube.py in _apply_spectral_function(arguments, outcube, function, **kwargs)
    108 
    109     if np.any(includemask):
--> 110         outcube[:,jj,ii] = function(spec, **kwargs)
    111     else:
[...]

Issue #514, opened 2018-12-13
Error returned from spectral_smooth in Python 3.7.0


ashleythomasbarnes:
I am trying to run some code I previously had working in Python 2.7.0 that used spectral-cube to do spectral smoothing (the code I am using is almost identical to that shown in the smoothing example: https://spectral-cube.readthedocs.io/en/latest/smoothing.html). However, I get the following error: 

```
AttributeError                            Traceback (most recent call last)
<ipython-input-5-db1df68caa65> in <module>()
     26 
     27 kernel = Gaussian1DKernel(gaussian_width)
---> 28 smcube = cube.spectral_smooth(kernel)

/anaconda3/lib/python3.7/site-packages/spectral_cube/spectral_cube.py in spectral_smooth(self, kernel, convolve, verbose, use_memmap, num_cores, **kwargs)
   2768                                                       use_memmap=use_memmap,
   2769                                                       verbose=verbose,
-> 2770                                                       **kwargs)
   2771 
   2772     def spectral_interpolate(self, spectral_grid,

/anaconda3/lib/python3.7/site-packages/spectral_cube/spectral_cube.py in _apply_function_parallel_spectral(self, function, data, num_cores, verbose, use_memmap, parallel, **kwargs)
   2719 
   2720             for arg in spectra:
-> 2721                 _apply_function(arg, outcube, function, **kwargs)
   2722                 pbu()
   2723 

/anaconda3/lib/python3.7/site-packages/spectral_cube/spectral_cube.py in _apply_function(arguments, outcube, function, **kwargs)
     98 
     99     if any(includemask):
--> 100         outcube[:,jj,ii] = function(spec, **kwargs)
    101     else:
    102         outcube[:,jj,ii] = spec

/anaconda3/lib/python3.7/site-packages/astropy/utils/decorators.py in convolve(array, kernel, boundary, fill_value, nan_treatment, normalize_kernel, mask, preserve_nan, normalization_zero_tol)
    840             name = func.__name__
    841 
--> 842         func = make_function_with_signature(func, name=name, **wrapped_args)
    843         func = functools.update_wrapper(func, wrapped, assigned=assigned,
    844                                         updated=updated)

/anaconda3/lib/python3.7/site-packages/astropy/nddata/decorators.py in wrapper(data, *args, **kwargs)
    243                                   AstropyUserWarning)
    244 
--> 245             result = func(data, *args, **kwargs)
    246 
    247             if unpack and repack:

[...]

PR #578, opened 2019-09-17
Catch kernels defined as a quantity

e-koch:
This is a lazy fix for #514 and #577. A `UnitsError` (maybe a different error would be appropriate) is raised when a kernel is defined as a quantity, which runs into issues with the parallel operations that use a unitless version of the cube.


e-koch:
@keflavich -- This looks like a new issue just for the memmap output: https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/585940799#L1537

keflavich:
Darn.  I don't immediately see how to fix that, I'll have to look deeper.

Issue #580, opened 2019-09-26
I broke the build by merging something green


keflavich:
We are on a "broken master" caused by upstream changes.  It looks like there are several independent, simultaneous failures.

https://travis-ci.org/radio-astro-tools/spectral-cube/builds/590110581

* [ ] python2.7 + numpy 1.17 won't build.   https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/590110587#L877
* [ ] reproject doesn't take independent_celestial_slices (https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/590110589#L2039): is this a version issue?
* [ ] Some totally wacky message (https://travis-ci.org/radio-astro-tools/spectral-cube/jobs/590110588#L1404) that probably points back to incompatibilities:
ERROR: reproject 0.6.dev0 has requirement astropy>=3.2, but you'll have astropy 2.0.14 which is incompatible.
ERROR: reproject 0.6.dev0 has requirement numpy>=1.13, but you'll have numpy 1.11.3 which is incompatible.

I'm going to try dropping _testing_ for old things (py2, astropy-LTS), so we're "dropping support" for these, but I still want them to work...

astrofrog:
The independent_celestial_slices argument is no longer needed but I should add a deprecation warning to reproject!

astrofrog:
Also reproject dev requires astropy 3.2 - of course you can still support old versions of reproject but if you are testing against reproject dev you’ll need astropy 3.2+

astrofrog:
Just to clarify, what I mean is that I don’t think you should have to drop testing for old versions - just don’t rely on reproject dev for those builds

PR #581, opened 2019-09-26
Fix for Issue #580

keflavich:
Everything broke, this is to try to get tests green again

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2553,6 +2553,10 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         if LooseVersion(version) < "0.5":
             raise Warning("Requires version >=0.5 of reproject. The current "
                           "version is: {}".format(version))
+        elif LooseVersion(version) >= 0.6:
+            reproj_kwargs = {}
+        else:
+            reproj_kwargs = {'independent_celestial_slices': True}
 
         from reproject import reproject_interp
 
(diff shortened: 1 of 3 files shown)


keflavich:
#580

keflavich:
@astrofrog re: https://github.com/astropy/reproject/pull/166#issuecomment-535939570 see https://github.com/radio-astro-tools/spectral-cube/pull/581/commits/629860a77a66b0a2392a44a0fae513041574b781

keflavich:
The remaining issues are some interaction w/numpy 1.17 and astropy

keflavich:
@e-koch this should fix the outstanding issues that were holding back your other PRs, so with your approval, let's merge this and then get your others merged.

spectral_cube/spectral_cube.py
```diff
@@ -2563,25 +2567,30 @@ def reproject(self, header, order='bilinear', use_memmap=False,
         shape_out = tuple([header['NAXIS{0}'.format(i + 1)] for i in
                            range(header['NAXIS'])][::-1])
 
+        if filled:
+            data = self.filled_data[:]
+        else:
+            data = self._data
```
e-koch: Is there a unit difference between `_data` and `filled_data`? Does it matter if the output is written to a memmap? This hasn't been a issue and is the default setting so maybe it doesn't matter.

keflavich: huh, should probbably be `unitless_filled_data` here....

keflavich: the units should be the same & inherited from the cube

Issue #588, opened 2019-11-21
SpectralCube.reproject does not properly update headers of resulting image


aakepley:
When I reproject a cube, it does not appear to update the header with the new WCS. I've put a minimal working example before. The headers for regridCube and otherCube are identical. My assumption was that the regridded CRPIX* etc in the  header would match that of the base cube. See below for an example.

```
# Now let's test my fix for the reprojection in spectral cube
from spectral_cube import SpectralCube

baseCubeFits = 'NGC2903_HCN_rebase3_smooth1.3_hanning1.new_smooth.fits'   
baseCube =SpectralCube.read(baseCubeFits)   

otherDataFits='NGC2903_heracles_gauss15_fixedheader.fits' 
otherCube = SpectralCube.read(otherDataFits) 

regridCube = otherCube.spectral_interpolate(baseCube.spectral_axis)
regridCube.reproject(baseCube.header)

regridCube.write('test_regridcube.fits',overwrite=True)

baseCube.header
# NAXIS   =                    3                                                  
# NAXIS1  =                   85                                                  
# NAXIS2  =                   86                                                  
# NAXIS3  =                  132           
# CRPIX1  =                 71.5 / Pixel coordinate of reference point            
# CRPIX2  =                 32.5 / Pixel coordinate of reference point            
# CRPIX3  =               -28.75 / Pixel coordinate of reference point            
# CDELT1  = -0.00057096828855099 / [deg] Coordinate increment at reference point  
# CDELT2  =  0.00057096828855099 / [deg] Coordinate increment at reference point  
# CDELT3  =    9.909510480999799 / [m/s] Coordinate increment at reference point  
# CUNIT1  = 'deg'                / Units of coordinate increment and value        
# CUNIT2  = 'deg'                / Units of coordinate increment and value        
# CUNIT3  = 'km s-1  '           / Units of coordinate increment and value        
# CTYPE1  = 'RA---TAN'           / Right ascension, gnomonic projection           
# CTYPE2  = 'DEC--TAN'           / Declination, gnomonic projection               
# CTYPE3  = 'VRAD'               / Radio velocity (linear)                        
# CRVAL1  =       143.0246014674 / [deg] Coordinate value at reference point      
# CRVAL2  =      21.495097040103 / [deg] Coordinate value at reference point      
# CRVAL3  =     -389.27897335034 / [m/s] Coordinate value at reference point   

otherCube.header
# NAXIS   =                    3                                                  
[...]

PR #589, opened 2019-11-21
Test equivalence of reprojected cube WCS

e-koch:
reproject tests were not checking the output WCS. See #588, though I wasn't able to reproduce the error locally.


e-koch:
WCS checks are fine.

docs build is failing when building wheels for regions. Needs np>1.16

keflavich:
Awesome, thanks!

PR #592, opened 2019-12-07
Create array-like reader for CASA images so we can retrieve just the chunks we need

keflavich:
Next step is to daskify this.   Also, I've only implemented the minimal required attributes for daskification, not for actual use as a spectral-cube class, afaik.

WIP!!


keflavich:
This is ready to merge if & when tests pass.

PR #598, opened 2020-01-13
Use fixtures to define test data instead of writing test files to the source/installed tree

astrofrog:
I've been working on refactoring the infrastructure of spectral-cube to follow best practices in Python packaging. This specific PR replaces the ``make_test_data.py`` script which needed to be run manually to generate the test FITS files with pytest fixtures, which write the test files to temporary directories.

A few reference results have changed because of the changes in random number generation. I believe the current approach is more stable going forward though - before this PR, the data files were generated sequentially and the order mattered for the random number generation. With this PR, each test data file is self contained and has its own random number seed set.

This will also make it possible to properly run the tests with ``spectral_cube.test()`` for installed versions of the package.


astrofrog:
In the docs, I switched from using ``adv.fits`` to ``example_cube.fits`` which is bundled by default with the package.

keflavich:
Approved pending tests passing.

astrofrog:
Tests are passing so I'll go ahead and merge - thanks for the review @keflavich!

PR #595, opened 2019-12-08
CASA chunk-based reading fixes

keflavich:
PR #592 was riddled with errors we totally missed because our casa image test coverage is nonexistent.  This will eventually fix that...

PR #600, opened 2020-01-15
Update I/O infrastructure to use Astropy unified I/O

astrofrog:
This updates the I/O in spectral-cube to use the astropy I/O registry machinery. This should in principle make it easier to add new formats in future, and avoids having to have hard-coded if statements to check what format is being used.

At the moment, there is a bit of code/docstring duplication which I hope to get rid of, so this is a work in progress (but I'm opening this to make sure the CI passes as-is already). Note that if we ever switch to inheriting from NDData, a whole bunch of code in ``io.core`` will no longer be needed.

This also fixes the tox configuration which was previously not properly installing the developer version of astropy, and fixes some related issues.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 
 - Refactor package infrastructure to no longer use astropy-helpers. #599
 
+- Switch to using unified I/O infrastructure from Astropy. #600
+
 0.4.5 (unreleased)
 ------------------
  - Added support for casatools-based io in #541 and beam reading from CASA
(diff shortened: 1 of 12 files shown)


keflavich:
Looks good afaict.  Would it be worth bringing in an external reviewer to check your work, or shall we just trust the tests?

astrofrog:
I'm not sure if a *code* review is going to be very easy since there's a bunch of moving code around, so to some extent maybe trusting the tests is fine, and we can then add regression tests if issues come up? I just pushed another commit that tries to avoid some code duplication.

Having said this, we might want to consider merging https://github.com/radio-astro-tools/spectral-cube/pull/595 first and rebasing this PR, so that we properly test for dangling file handles (since this PR touches code related to files and file handles).

astrofrog:
I've now rebased, so let's see if any new failures come up

astrofrog:
@keflavich - looks like this is good to go

PR #603, opened 2020-01-16
Fix bug when using StokesSpectralCube.read

astrofrog:
This fixes a bug which caused StokesSpectralCube.read to not work correctly following my refactoring of the I/O in #600


astrofrog:
This now also fixes a bug with the auto-identification of CASA images. I had to include a workaround for a bug in astropy core related to when the filename is for a directory - I'll submit a patch to astropy core but in the mean time things should work here with the ``StringWrapper`` workaround.

astrofrog:
And finally this fixes StokesSpectralCube.read for CASA data - and overall improves the test coverage.

Issue #605, opened 2020-01-28
Write our own CASA file reader


keflavich:
Some experiments w/CASA file reading have been utter disasters - multiple different classes of traceback followed by reproducible but different each time segmentation faults.  CASA is clearly the problem and is providing a barrier to progress with `.image` file types.

CASA images are simply binary files laid out in a somewhat-reasonable way on disk.  

We already use `tb.getdminfo` to get the array layout:
```
dminfo = tb.getdminfo()
{'*1': {'COLUMNS': array(['map'],
        dtype='|S4'),
  'NAME': 'map',
  'SEQNR': 0,
  'SPEC': {'DEFAULTTILESHAPE': array([  8,  32, 128], dtype=int32),
   'HYPERCUBES': {'*1': {'BucketSize': 131072,
     'CellShape': array([ 16,  64, 256], dtype=int32),
     'CubeShape': array([ 16,  64, 256], dtype=int32),
     'ID': {},
     'TileShape': array([  8,  32, 128], dtype=int32)}},
   'MAXIMUMCACHESIZE': 0,
   'MaxCacheSize': 0,
   'SEQNR': 0},
  'TYPE': 'TiledCellStMan'}}
```

So we can do something like:
```
chunksize = dminfo['*1']['SPEC']['DEFAULTTILESHAPE']
totalsize = dminfo['*1']['SPEC']['HYPERCUBES']['*1']['CubeShape']
fh = open('test.image/table.f0_TSM0', 'rb')
chunks = [np.fromfile(fh, dtype='float32', count=np.product(chunksize)).reshape(chunksize, order='F') for ii in range(np.product(totalsize) / np.product(chunksize))]
```
to load the chunks.  Then there's more work to do to figure out how to stitch these together properly (in a dask reader).  Note that the chunks could readily be constructed as `mmap`s instead of using `fromfile`.

keflavich:
(it's unclear what happens in the case of odd-numbered chunks, etc., too)

keflavich:
This might work:
```
def casa_image_reader(imagename):

    tb.open(imagename)
    dminfo = tb.getdminfo()
    tb.close()

    chunksize = dminfo['*1']['SPEC']['DEFAULTTILESHAPE']
    totalsize = dminfo['*1']['SPEC']['HYPERCUBES']['*1']['CubeShape']
    stacks = totalsize / chunksize
    nchunks = np.product(totalsize) // np.product(chunksize)
[...]

PR #607, opened 2020-02-09
Add an experimental "homebrew" CASA .image reader

keflavich:
Partial solution to #605.  I think this code now clearly-ish outlines what is needed to build an array from the bytes on disk; the next step is to daskify this reader process.

Also, I want a large series of tests to actually run.  This is just what I did to get to the "good enough" stage.  We really need to test this on different:

 * 1. numbers of dimensions (1,2,3,4,5)
 * 2. numbers of chunks (<=1, 2-10)


keflavich:
@astrofrog maybe give this a quick review?  This might be the best solution (once dasked) to all of the CASA segfaulty problems

keflavich:
so I definitely have this mostly-solved, but the steps of not-loading-into-memory and daskifying the reader are still a step beyond me.  Creating our own slice, as we did with `ArrayLikeCasaData`, will definitely work, but it would require manual concatenation, etc.  There has to be a better, simpler way.

In the latest commit, I added some commented-out code that shows how to get an n-dimensional array of the chunk indices.  How you populate an n-dimensional array-like object _without_ reading every byte is the part I haven't solved.  It's probably trivial, though, and I'm just overlooking a single, simple function.

spectral_cube/io/casa_image.py
```diff
@@ -355,6 +355,60 @@ def load_casa_image(filename, skipdata=False,
...
+        remaining_dims = rstacks
+        if len(remaining_dims) == 0:
+            assert kk == 0
+            rslt = np.concatenate(rslt, 0)
+        else:
+            cut = np.product(remaining_dims)
+            assert cut % 1 == 0
+            cut = int(cut)
+            rslt = [np.concatenate(rslt[ii::cut], kk) for ii in range(cut)]
+        jj += 1
```
astrofrog: A more memory efficient way to do this which doesn't rely on too many temporary arrays as this does would be to just create the final array then use index to insert each chunk at the correct location.

spectral_cube/tests/test_casafuncs.py
```diff
@@ -96,6 +107,45 @@ def test_casa_read_stokes(data_advs, tmp_path):
...
+
+@pytest.mark.skipif(not casaOK, reason='CASA tests must be run in a CASA environment.')
+@pytest.mark.parametrize('shape', ((129,128,130), (513,128,128), (128,513,128),
+                                   (128,128,513), (512,64,64)),
+                         )
+def test_casa_numpyreader_read_shape(shape, tmp_path):
+
+    cube = SpectralCube.read(filename)
+
+    make_casa_testimage_of_shape(shape, tmp_path / 'casa.image')
```
astrofrog: Do you have any control over the chunking? If so would be good to try different chunking options?

spectral_cube/io/casa_image.py
astrofrog: And just saw the comments below :)

spectral_cube/tests/test_casafuncs.py
keflavich: no, the chunking is on-disk

spectral_cube/io/casa_image.py
keflavich: Yep.  I am irrationally confident that there is a trivial way to do what I'm trying to do in a single line, but I haven't figured out what that magical invocation is yet.

spectral_cube/tests/test_casafuncs.py
astrofrog: just to clarify, I mean can you create files on disk with different degrees of chunking?

keflavich: afaik, no.  CASA makes some choice for you, and I do not know of any way to control that.  It's at least not obviously accessible in any of the current interfaces.

astrofrog:
@keflavich - I've done some more work to improve performance, and have switched the CASA loader to use the dask reader by default. There is still some work to do, in particular in relation to masks. For now I have modified the code so that we can read in the data for the mask, except that (a) the dtype is probably wrong for datasets with masks, and (b) a lot of datasets I'm looking at have a very small mask file which must be effectively empty, so we need to handle this case.

astrofrog:
I've now managed to properly read in the masks (which are actually stored as sequences of bits, so this requires a different approach to the data), and I've also fixed it for 64-bit floating point data. I have all the tests passing locally. We can also read in different masks by name with the low-level function, and that could potentially be exposed from the SpectralCube reader.

At this point, the remaining to-dos are:

* [ ] Clean up the CASA tests - I added a new detailed test of the low-level function, but I need to check if we can remove or combine some of the other tests
* [ ] Make sure that the SpectralCube CASA reader (not the low-level function) is efficient memory wise and doesn't try and cause all the data to be read in

Otherwise I think this is close to ready. Note that for now I haven't seen performance benefits from using dask, in part due to the small chunk size, but what's important here is first to make sure things are correct and efficient memory-wise, and don't segfault!

astrofrog:
@keflavich - I think this is ready for review. One thing that I'd like to understand is how CASA chooses chunk sizes and if we can know for sure they will always be smaller than a certain size. If so, then we could replace the memmap with a fromfile since otherwise we have two 'layers' of lazy-loading - the dask chunks and the memmap within that, which maybe isn't needed. However, if chunks can be arbitrarily large, we should keep the memmap. In any case, this is not something that needs to hold back this PR. As-is, this PR seems to work and the tests pass. I've tried to make the internal dask reader as fast as possible but it can still take ~5 seconds for the data and mask each to construct the dask array when there are ~30,000 chunks.

spectral_cube/io/casa_image.py
```diff
@@ -356,6 +255,173 @@ def load_casa_image(filename, skipdata=False,
...
+        self._kwargs = kwargs
+        self.shape = kwargs['shape'][::-1]
+        self.dtype = kwargs['dtype']
+        self.ndim = len(self.shape)
+
+    def __getitem__(self, item):
+        # We open a file manually and return an in-memory copy of the array
+        # otherwise the file doesn't get closed properly.
+        with open(self._filename) as f:
+            return np.memmap(f, mode='readonly', order='F', **self._kwargs).T[item].copy()
```
keflavich: This is kind of a mess, but I see why it's needed.

There must be some way to define a `np.memmap`-like wrapper around `mmap.mmap`, the lower-level python function, that internally translates the slice syntax to the appropriate locations.  I'm not sure whether that would be more performant, but it would at least prevent the many-open-files problems.

astrofrog: Yes that could be a solution, I'll investigate whether it helpers with performance.

PR #608, opened 2020-02-20
Stokes I empty mask

e-koch:
Found a small bug in checking the Stokes I mask from #607.

### DIFF
--- a/spectral_cube/io/casa_image.py
+++ b/spectral_cube/io/casa_image.py
@@ -244,7 +244,7 @@ def load_casa_image(filename, skipdata=False,
 
 
         cube = StokesSpectralCube(stokes_data=data)
-        if mask is not None:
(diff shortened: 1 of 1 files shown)


e-koch:
I'm unsure how common it is for the casa image to have no mask. The bug came up when reading in some temporary images that are forced to have no mask.

e-koch:
@astrofrog -- I've had trouble making a test for this. `ia.fromfits` and force creating an empty image (`ia.newimagefromshape`) both automatically create a mask. It seems like this won't be a common case.

astrofrog:
@e-koch - the easiest way might be to generate a dataset then remove the mask in the test before trying to parse it?

astrofrog:
I cherry-picked the commit here into #609 and added a regression test

PR #609, opened 2020-02-21
Added Python and Numpy implementation of getdminfo

astrofrog:
This is still a work in progress, but is close to done. Remaining to-dos:

* [x] Switch the CASA dask reader to use this new function
* [x] Fix test for failing cases - there are some corner cases where the bucket size isn't being read in/computed correctly
* [x] Check that there are no other failures in CI

@keflavich - this brings us one step closer to being able to read in CASA datasets without CASA - however before that can happen I would need to add Python implementations of:

```
    casa_cs = ia.coordsys()
    unit = ia.brightnessunit()
    beam_ = ia.restoringbeam()
```

Given that it wasn't too hard to do ``getdminfo``, it might not be too hard to do, though I can see the coordinate system being a little bit trickier potentially.

In any case, with this PR, ``casa_image_dask_reader`` is now CASA-free - it's mainly the meta-data that still needs CASA.

PR #613, opened 2020-02-27
No longer require CASA for coordsys to WCS conversion

astrofrog:
This is a work in progress which makes it so that we no longer need CASA to convert the CASA coordsys to Astropy WCS. Remaining to-dos:

* [x] Add changelog entry
* [x] Try and add some more test cases
* [x] Add support for the 'linear' axis type, which is relevant if no WCS is set
* [x] Check casacore code for other supported spectral systems (there are a few in ``FITSCoordinateUtil.cc``)
* [x] Determine why CDELT disagrees at the 1e-5 relative level for spectral axis (see #614)
* [x] Rebase once https://github.com/radio-astro-tools/spectral-cube/pull/609 is merged

With this PR it's now possible to fully load CASA spectral cubes without CASA!

### DIFF
--- a/spectral_cube/io/casa_wcs.py
+++ b/spectral_cube/io/casa_wcs.py
@@ -47,6 +47,8 @@ for system in EQUATORIAL_SYSTEMS:
 def sanitize_unit(unit):
     if unit == "'":
         return 'arcmin'
+    elif unit == '"':
+        return 'arcsec'
     else:
         return unit
 
@@ -114,13 +116,18 @@ def wcs_casa2astropy(coordsys):
 
     for coord_type in ('direction', 'spectral', 'stokes', 'linear'):
 
+        # Each coordinate type is stored with a numerical index,
+        # so for example a cube might have direction0 and
+        # spectral1, or spectral0, direction1, stokes2. The actual
+        # index is irrelevant for the rest of the loop.
         for index in range(len(worldmap)):
             if f'{coord_type}{index}' in coordsys:
+                data = coordsys[f'{coord_type}{index}']
                 break
         else:
+            # Skip the rest of the loop for the current coord_type
+            # since the coord_type wasn't found with any index.
             continue
-
-        data = coordsys[f'{coord_type}{index}']
 
         if coord_type == 'direction':
             idx1, idx2 = worldmap[index] + 1
--- a/spectral_cube/io/casa_wcs.py
+++ b/spectral_cube/io/casa_wcs.py
@@ -6,6 +6,13 @@ from astropy.time import Time
 __all__ = ['wcs_casa2astropy']
 
 
+def sanitize_unit(unit):
+    if unit == "'":
+        return 'arcmin'
+    else:
+        return unit
+
+
 def wcs_casa2astropy(coordsys):
     """
     Convert a casac.coordsys object into an astropy.wcs.WCS object
@@ -14,6 +21,9 @@ def wcs_casa2astropy(coordsys):
     # Rather than try and parse the CASA coordsys ourselves, we delegate
     # to CASA by getting it to write out a FITS file and reading back in
     # using WCS
+
+    from pprint import pprint
+    pprint(coordsys)
 
     header = fits.Header()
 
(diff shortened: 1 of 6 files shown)


astrofrog:
@e-koch - could you try out this PR? Note that you could try it in an environment without CASA since CASA should no longer be required to read in cubes with ``SpectralCube.read``

astrofrog:
(note that this is not worth reviewing line by line until #609 is merged, since this PR includes the changes from there)

e-koch:
@astrofrog - Everything seems to be working using the same cubes I used testing #609!

astrofrog:
@e-koch - great!

spectral_cube/io/casa_wcs.py
```diff
@@ -1,11 +1,57 @@
...
+}
+
+for system in EQUATORIAL_SYSTEMS:
+    AXES_TO_CTYPE[system] = {'Right Ascension': 'RA--',
+                             'Declination': 'DEC-'}
+
+
+def sanitize_unit(unit):
+    if unit == "'":
+        return 'arcmin'
```
keflavich: do we need `"` for arcsec, or does that not appear?

spectral_cube/io/casa_wcs.py
```diff
@@ -14,24 +60,142 @@ def wcs_casa2astropy(ndim, coordsys):
...
+            worldmap[index] = value
+
+    # Now iterate through the different coordinate types to populate the WCS
+
+    header['WCSAXES'] = np.max([np.max(idx) + 1 for idx in worldmap.values()])
+
+    # Initialize PC
+    for i in range(header['WCSAXES']):
+        for j in range(header['WCSAXES']):
+            header[f'PC{i+1}_{j+1}'] = 0.
```
keflavich:
```suggestion
    for ii in range(header['WCSAXES']):
        for jj in range(header['WCSAXES']):
            header[f'PC{ii+1}_{jj+1}'] = 0.
```

spectral_cube/io/casa_wcs.py
```diff
@@ -14,24 +60,142 @@ def wcs_casa2astropy(ndim, coordsys):
...
+
+    for coord_type in ('direction', 'spectral', 'stokes', 'linear'):
+
+        for index in range(len(worldmap)):
+            if f'{coord_type}{index}' in coordsys:
+                break
+        else:
+            continue
```
keflavich: Just commenting that I have never before seen the `for/else` used in practice.    It's uncommon enough that I would suggest including a comment about it; at first I was convinced this was a typo.

spectral_cube/io/casa_wcs.py
```diff
@@ -14,24 +60,142 @@ def wcs_casa2astropy(ndim, coordsys):
...
+                header[f'PV{idx2}_2'] = 0.
+            header[f'PC{idx1}_{idx1}'] = data['pc'][0, 0]
+            header[f'PC{idx1}_{idx2}'] = data['pc'][0, 1]
+            header[f'PC{idx2}_{idx1}'] = data['pc'][1, 0]
+            header[f'PC{idx2}_{idx2}'] = data['pc'][1, 1]
+        elif coord_type == 'stokes':
+            idx = worldmap[index][0] + 1
+            header[f'CTYPE{idx}'] = 'STOKES'
+            header[f'CRVAL{idx}'] = data['crval'][0]
+            header[f'CRPIX{idx}'] = data['crpix'][0] + 1
```
keflavich: so... this means CASA stores CRPIX in zero-indexed, not one-indexed form?

keflavich: (this comment applies elsewhere too - I'm just observing, I don't think any code change or even commentary is needed)

spectral_cube/io/casa_wcs.py
astrofrog: I haven't managed to find files that use this, but no harm in adding it, so I've done this.

spectral_cube/io/casa_wcs.py
astrofrog: I've added a comment. The ``else`` for a ``for`` loop is quite useful for situations like this!

spectral_cube/io/casa_wcs.py
astrofrog: Yes it seems so, which would make sense internally for calculations.

PR #616, opened 2020-03-02
Improvements to testing of CASA reader

astrofrog:
This does several things:

* Adds a simple CASA dataset in the repo
* Adds a test that uses that dataset which can then run in all CI configurations without CASA installed (currently all the CASA tests still require CASA to *generate* the test images from FITS)
* Fixes a bug with accessing scalar values from SpectralCube when backed by a dask array
* Adds MacOS X and Windows builds to the CI
* Fixes a bug that prevented CASA images from being read on Windows (they can now!)
* Skips a few tests related to parallelization and memory usage on Windows for now


astrofrog:
There are some Windows failures that I will need to investigate when I am next at my desktop machine.

astrofrog:
This is ready for review! I will investigate the parallel tests on Windows separately, though those issues may end up being moot if we switch to dask in future so maybe not high priority.

astrofrog:
Note that for now Astropy dev is required on Windows for the CASA I/O. Once Astropy 4.0.1 is out (with the https://github.com/astropy/astropy/pull/9866 fix) Astropy stable can be used.

astrofrog:
No the issue was that I was reading the ints as unsigned ints and then converting them to signed ints, and Windows is picker about overflow errors. In fact it was an issue on all platforms but I think only resulted in a warning on Mac/Linux.

PR #620, opened 2020-04-09
 Fix bug in VaryingResolutionSpectralCube.mask_channels

astrofrog:
Fixes a bug that caused ``VaryingResolutionSpectralCube.mask_channels`` to not preserve existing mask

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -1132,7 +1132,7 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
 
         return result
 
-    def with_mask(self, mask, inherit_mask=True, wcs_tolerance=None, **kwargs):
+    def with_mask(self, mask, inherit_mask=True, wcs_tolerance=None):
         """
         Return a new SpectralCube instance that contains a composite mask of
         the current SpectralCube and the new ``mask``.  Values of the mask that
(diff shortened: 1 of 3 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -1132,7 +1132,7 @@ def percentile(self, q, axis=None, iterate_rays=False, **kwargs):
 
         return result
 
-    def with_mask(self, mask, inherit_mask=True, wcs_tolerance=None):
+    def with_mask(self, mask, inherit_mask=True, wcs_tolerance=None, **kwargs):
```
astrofrog, line 1135: I'm not sure if this is the best approach, since it adds **kwargs to the public API, but otherwise we'd need to repeat some code.

keflavich: I think that's fine, though it might result in docs headaches later...

spectral_cube/spectral_cube.py
```diff
@@ -3974,10 +3974,8 @@ def mask_channels(self, goodchannels):
         mask = BooleanArrayMask(goodchannels[:,None,None], self._wcs,
                                 shape=self._data.shape)
```
keflavich, line 3975: these two lines don't do anything now, do they?

astrofrog: Removed

astrofrog:
@keflavich - I tried a different approach to avoid adding kwargs to with_mask, instead modifying goodbeams_mask on the final cube.

spectral_cube/spectral_cube.py
astrofrog: Removed - I've tried a different way now

Issue #627, opened 2020-04-28
Provide a way to temporarily store the data array in ``DaskSpectralCube`` to a zarr array


astrofrog:
We should provide a way in ``DaskSpectralCube`` to dump the current dask array to disk in a zarr array and then re-loading it straight away, as a way to effectively force a computation and avoid long chains of e.g. spectral then image then spectral operations which could be quite inefficient. This would be the equivalent of the ``use_memmap`` option in some methods in the plain ``SpectalCube``. Note however that this would just be a generic method rather than adding an option to all methods (alternatively it could be added to all methods via a decorator, but I think a separate explicit method would be nicer). Any suggestions for API?

keflavich:
Can we have a subclass, or a boolean property like `allow_huge=True/False` that does all operations through `zarr`-dumps when set?  So if you're using a really big cube, all operations happen through the non-memory path?

Otherwise, the optional argument to each operation sounds nice as an API, though a pain to implement.

PR #629, opened 2020-05-13
Save to temporary array and efficient spatial/spectral chunking

astrofrog:
This does a few separate things:

* Provides a way to force computations and dump the intermediate results to disk in a temporary zarr directory. This is done via a ``save_to_tmp_dir`` kwarg in each relevant method. Fixes https://github.com/radio-astro-tools/spectral-cube/issues/627
* Provides a way in ``apply_function_parallel_spectral/spatial`` to specify ``allow_chunks=True``, which can lead to performance improvements
* Adds documentation about these features, including a simple benchmark to demonstrate the benefits of ``allow_chunks``
* Adds a section about the performance benefits of the dask classes compared to the regular classes.

One thing that was bugging me is the ``parallel`` in ``apply_function_parallel_?`` - since whether it's parallel or not depends on the scheduler, and other methods can be parallel too, should we drop the ``parallel`` from the method names in the dask class? I don't really mind either way, but just wanted to raise this.

### DIFF
--- a/docs/dask.rst
+++ b/docs/dask.rst
@@ -183,7 +183,8 @@ to disk::
 
 The ``sigma`` argument is passed to the ``sigma_clip_with_nan`` function. We now call this
 again but specifying that the ``sigma_clip_with_nan`` function can also take cubes, using
-the ``accepts_chunks=True`` option::
+the ``accepts_chunks=True`` option (note that for this to work properly, the wrapped function
+needs to include ``axis=0`` in the call to :func:`~astropy.stats.sigma_clip` as shown above)::
 
     >>> clipped_cube = large.apply_function_parallel_spectral(sigma_clip_with_nan, sigma=3,
     ...                                                       accepts_chunks=True,
--- a/docs/dask.rst
+++ b/docs/dask.rst
@@ -84,7 +84,7 @@ re-read it immediately from disk (for users interested in how the data is stored
      n_y:      4  type_y: DEC--ARC  unit_y: deg    range:    31.243639 deg:   31.243739 deg
      n_s:      7  type_s: VRAD      unit_s: m / s  range:    14322.821 m / s:   14944.909 m / s
 
-Note that this requires the `zarr` and `fsspec <https://pypi.org/project/fsspec/>`_ packages to be
+Note that this requires the `zarr`_ and `fsspec <https://pypi.org/project/fsspec/>`_ packages to be
 installed.
 
 This can also be beneficial if you are using multiprocessing or multithreading to carry out calculations,
@@ -97,7 +97,7 @@ Like the :class:`~spectral_cube.SpectralCube` class, the
 :class:`~spectral_cube.DaskSpectralCube` and
 :class:`~spectral_cube.DaskVaryingResolutionSpectralCube` classes have methods for applying custom
 functions to all the spectra or all the spatial images in a cube:
-:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectra` and
+:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectral` and
 :meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spatial`. By default, these methods
 take functions that apply to individual spectra or images, but this can be quite slow for large
 spectral cubes. If possible, you should consider supplying a function that can accept 3-d cubes
@@ -116,7 +116,7 @@ the image plane::
 As an example, we will apply sigma clipping to all spectra in the cube. Note that there is a method
 to do this (:meth:`~spectral_cube.DaskSpectralCube.sigma_clip_spectrally`) but for the purposes of
 demonstration, we will set up the function ourselves and apply it with
-:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectra`. We will use the
+:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectral`. We will use the
 :func:`~astropy.stats.sigma_clip` function from astropy::
 
     >>> from astropy.stats import sigma_clip
@@ -133,7 +133,7 @@ silence, so we can do this here too::
     ...         warnings.simplefilter('ignore')
     ...         return sigma_clip(*args, **kwargs).filled(np.nan)
 
-Let's now call :meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectra`, including the
+Let's now call :meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectral`, including the
 ``save_to_tmp_dir`` option mentioned previously to force the calculation and the storage of the result
 to disk::
 
--- a/docs/dask.rst
+++ b/docs/dask.rst
@@ -1,6 +1,9 @@
 Integration with dask
 =====================
 
+Getting started
+---------------
+
 When loading a cube with the :class:`~spectral_cube.SpectralCube` class, it is possible to optionally
 specify the ``use_dask`` keyword argument to control whether or not to use new experimental classes
 (:class:`~spectral_cube.DaskSpectralCube` and :class:`~spectral_cube.DaskVaryingResolutionSpectralCube`)
@@ -21,6 +24,9 @@ To read in a FITS cube using the dask-enabled classes, you can do::
 
 Most of the properties and methods that normally work with :class:`~spectral_cube.SpectralCube`
 should continue to work with :class:`~spectral_cube.DaskSpectralCube`.
+
+Schedulers and parallel computations
+------------------------------------
 
 By default, we use the ``'synchronous'`` `dask scheduler <https://docs.dask.org/en/latest/scheduler-overview.html>`_
 which means that calculations are run in a single process/thread. However, you can control this using the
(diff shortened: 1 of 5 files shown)


docs/dask.rst
```diff
@@ -53,3 +62,132 @@ a progress bar:
...
+    >>> cube_new = cube.sigma_clip_spectrally(3, save_to_tmp_dir=True)  # doctest: +IGNORE_OUTPUT
+    [########################################] | 100% Completed |  0.1s
+    >>> cube_new
+    DaskSpectralCube with shape=(7, 4, 3) and unit=Jy / beam:
+     n_x:      3  type_x: RA---ARC  unit_x: deg    range:    52.231466 deg:   52.231544 deg
+     n_y:      4  type_y: DEC--ARC  unit_y: deg    range:    31.243639 deg:   31.243739 deg
+     n_s:      7  type_s: VRAD      unit_s: m / s  range:    14322.821 m / s:   14944.909 m / s
+
+Note that this requires the `zarr`_ and `fsspec <https://pypi.org/project/fsspec/>`_ packages to be
+installed.
```
keflavich, line 90: We need to add something about this to the `installation` section

docs/dask.rst
```diff
@@ -53,3 +62,132 @@ a progress bar:
...
+:class:`~spectral_cube.DaskSpectralCube` and
+:class:`~spectral_cube.DaskVaryingResolutionSpectralCube` classes have methods for applying custom
+functions to all the spectra or all the spatial images in a cube:
+:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectral` and
+:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spatial`. By default, these methods
+take functions that apply to individual spectra or images, but this can be quite slow for large
+spectral cubes. If possible, you should consider supplying a function that can accept 3-d cubes
+and operate on all spectra or image slices in a vectorized way.
+
+To demonstrate this, we will read in a large CASA dataset with 623 channels and 768x768 pixels in
```
keflavich, line 144:
```suggestion
To demonstrate this, we will read in a mid-sized CASA dataset with 623 channels and 768x768 pixels in
```

docs/dask.rst
```diff
@@ -53,3 +62,132 @@ a progress bar:
...
+    >>> large = SpectralCube.read('large_spectral_cube.image', format='casa_image', use_dask=True)  # doctest: +SKIP
+    >>> large  # doctest: +SKIP
+    DaskVaryingResolutionSpectralCube with shape=(623, 768, 768) and unit=Jy / beam:
+    n_x:    768  type_x: RA---SIN  unit_x: deg    range:   290.899389 deg:  290.932404 deg
+    n_y:    768  type_y: DEC--SIN  unit_y: deg    range:    14.501466 deg:   14.533425 deg
+    n_s:    623  type_s: FREQ      unit_s: Hz     range: 216201517973.483 Hz:216277445708.200 Hz
+
+As an example, we will apply sigma clipping to all spectra in the cube. Note that there is a method
+to do this (:meth:`~spectral_cube.DaskSpectralCube.sigma_clip_spectrally`) but for the purposes of
+demonstration, we will set up the function ourselves and apply it with
```
keflavich, line 156: This is great, we should definitely have examples like this!

docs/dask.rst
```diff
@@ -53,3 +62,132 @@ a progress bar:
...
+:meth:`~spectral_cube.DaskSpectralCube.apply_function_parallel_spectral`. We will use the
+:func:`~astropy.stats.sigma_clip` function from astropy::
+
+    >>> from astropy.stats import sigma_clip
+
+By default, this function returns masked arrays, but to apply this to our
+spectral cube, we need it to return a plain Numpy array with NaNs for the masked
+values. In addition, the original function tends to return warnings which we want to
```
keflavich, line 164:
```suggestion
values. In addition, the original function tends to return warnings we want to
```

docs/dask.rst
```diff
@@ -53,3 +62,132 @@ a progress bar:
...
+``save_to_tmp_dir`` option mentioned previously to force the calculation and the storage of the result
+to disk::
+
+    >>> clipped_cube = large.apply_function_parallel_spectral(sigma_clip_with_nan, sigma=3,
+    ...                                                       save_to_tmp_dir=True)  # doctest: +SKIP
+    [########################################] | 100% Completed |  1min 42.3s
+
+The ``sigma`` argument is passed to the ``sigma_clip_with_nan`` function. We now call this
+again but specifying that the ``sigma_clip_with_nan`` function can also take cubes, using
+the ``accepts_chunks=True`` option::
```
keflavich, line 186: is this true?  I thought `sigma_clip` would change behavior if not specifically run on a spectrum?  Have you verified the results are identical?

spectral_cube/dask_spectral_cube.py
```diff
@@ -58,6 +68,26 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def add_save_to_tmp_dir_option(function):
```
keflavich, line 71: should this also append the docstring, instead fo c&p'ing it below?

spectral_cube/dask_spectral_cube.py
```diff
@@ -422,24 +486,47 @@ def apply_function_parallel_spectral(self,
...
 
-        data = self._get_filled_data(fill=self._fill_value)
+        def wrapper(data, **kwargs):
+            if data.size > 0:
+                return function(data, **kwargs)
+            else:
+                return data
+
+        if accepts_chunks:
+
```
keflavich, line 510: this whitespace looks a bit off to me - not a big deal, but does this match the style used elsewhere?

spectral_cube/dask_spectral_cube.py
```diff
@@ -422,24 +486,47 @@ def apply_function_parallel_spectral(self,
...
+            return self._map_blocks_to_cube(wrapper,
+                                            rechunk=(-1, 'auto', 'auto'), **kwargs)
+
```
keflavich, line 513: ditto here

spectral_cube/dask_spectral_cube.py
```diff
@@ -422,24 +486,47 @@ def apply_function_parallel_spectral(self,
...
+        else:
+
```
keflavich, line 515: and here

spectral_cube/tests/test_dask.py
```diff
@@ -38,3 +40,12 @@ def test_scheduler(data_adv):
...
     cube._compute(fake_array)
     assert fake_array.kwargs == {'scheduler': 'threads'}
+
+
+def test_save_to_tmp_dir(data_adv):
+    pytest.importorskip('zarr')
+    cube = DaskSpectralCube.read(data_adv)
+    cube_new = cube.sigma_clip_spectrally(3, save_to_tmp_dir=True)
+    # The following test won't necessarily always work in future since the name
+    # is not really guaranteed, but this is pragrmatic enough for now
```
keflavich, line 50:
```suggestion
    # is not really guaranteed, but this is pragmatic enough for now
```

docs/dask.rst
astrofrog: The ``axis=0`` in the wrapped function is what makes it work ok

spectral_cube/dask_spectral_cube.py
astrofrog: Perhaps, but making it do it properly and insert it into the right section is a bit of a pain. It wasn't that hard to copy/paste ;)

docs/dask.rst
astrofrog: Done

docs/dask.rst
astrofrog: I've tried to clarify this

spectral_cube/dask_spectral_cube.py
astrofrog: Fixed

spectral_cube/dask_spectral_cube.py
astrofrog: Fixed

spectral_cube/dask_spectral_cube.py
astrofrog: Fixed

astrofrog:
@keflavich - I just pushed some fixes!

spectral_cube/dask_spectral_cube.py
keflavich: fair enough - I like the auto-add-docstring approach because it means there's a single point for maintenance, but it does make it harder to find the docs at not-runtime.

spectral_cube/dask_spectral_cube.py
```diff
@@ -422,24 +486,41 @@ def apply_function_parallel_spectral(self,
...
-                                   fill_value=self.fill_value)
+        if accepts_chunks:
+            return self._map_blocks_to_cube(wrapper,
+                                            rechunk=(-1, 'auto', 'auto'), **kwargs)
+        else:
+            data = self._get_filled_data(fill=self._fill_value)
+            # apply_along_axis returns an array with a single chunk, but we
+            # need to rechunk here to avoid issues when writing out the data
+            # even if it results in a poorer performance.
+            data = data.rechunk((-1, 50, 50))
```
e-koch, line 517: Is (50, 50) just a reasonable compromise for the write out? I'm wondering if there is any benefit to being able to change the spatial dim sizes.

docs/dask.rst
```diff
@@ -53,3 +62,133 @@ a progress bar:
...
+
+Saving intermediate results to disk
+-----------------------------------
+
+When calling methods such as for example :meth:`~spectral_cube.DaskSpectralCube.convolve_to` or any other
+methods that return a cube, the result is not immediately calculated - instead, the result is only computed
+when data is accessed directly (for example via `~spectral_cube.DaskSpectralCube.filled_data`), or when
+writing the cube to disk, for example as a FITS file. However, when doing several operations in a row, such
+as spectrally smoothing the cube, then spatially smoothing it, it can be more efficient to store intermediate
+results to disk. All methods that return a cube can therefore take the ``save_to_tmp_dir`` option which can
```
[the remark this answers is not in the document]
astrofrog: I'll add that in a follow-up PR.

docs/dask.rst
```diff
@@ -53,3 +62,133 @@ a progress bar:
...
+be set to `True` to compute the result of the operation immediately, save it to a temporary directory, and
+re-read it immediately from disk (for users interested in how the data is stored, it is stored as a
```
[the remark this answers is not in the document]
astrofrog: I think this might be system-dependent, I'm basically just using the Python ``tempfile`` module so I assume the system cleans this up from time to time (e.g. at reboot). I'll open an issue so that we can investigate in detail since these files will be big.

spectral_cube/dask_spectral_cube.py
astrofrog: It was a reasonable compromise for the datasets I tried, but I think I could perhaps write it more generically, will give it a shot.

docs/dask.rst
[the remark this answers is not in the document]
astrofrog: See https://github.com/radio-astro-tools/spectral-cube/issues/633

docs/dask.rst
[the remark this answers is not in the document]
astrofrog: See #632

spectral_cube/dask_spectral_cube.py
astrofrog: See #632

Issue #630, opened 2020-05-15
patch **kwargs passing in convolution


akleroy:
In convolution, convfunc is defined as a convenience function but drops the **kwarg passing. Line 3124 or so:

        def convfunc(img):
            return convolve(img, convolution_kernel, normalize_kernel=True,
                            **kwargs)

Can this be updated to:

        def convfunc(img, **kwargs):

please.

Thank you!

akleroy:
Or if I'm just screwing something up let me know and close the ticket.

keflavich:
Yes, make a pr!

keflavich:
So, the kwargs are already being passed to `convolve` directly, they're circumventing the intermediate `convfunc`.  I'm not sure there's any advantage - or harm - in the approach you've suggested.  Let's continue discussion on #631

e-koch:
Fixed in #631

PR #631, opened 2020-05-15
Update spectral_cube.py

akleroy:
I think this fixes the ability to pass kwargs to the convolution routine in convolve_to()

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1229,11 +1229,13 @@ class DaskSpectralCube(DaskSpectralCubeMixin, SpectralCube):
         convolution_kernel = beam.deconvolve(self.beam).as_kernel(pixscale)
         kernel = convolution_kernel.array.reshape((1,) + convolution_kernel.array.shape)
 
-        def convfunc(img):
+        # See #631: kwargs get passed within self.apply_function_parallel_spatial
+        def convfunc(img, **kwargs):
             return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape)
 
         return self.apply_function_parallel_spatial(convfunc,
-                                                    accepts_chunks=True).with_beam(beam)
+                                                    accepts_chunks=True,
+                                                    **kwargs).with_beam(beam)
 
 
 class DaskVaryingResolutionSpectralCube(DaskSpectralCubeMixin, VaryingResolutionSpectralCube):
@@ -1358,7 +1360,8 @@ class DaskVaryingResolutionSpectralCube(DaskSpectralCubeMixin, VaryingResolution
         beams = da.from_array(np.array(beams, dtype=np.object)
                               .reshape((len(beams), 1, 1)), chunks=(-1, -1, -1))
 
-        def convfunc(img, beam):
+        # See #631: kwargs get passed within self.apply_function_parallel_spatial
+        def convfunc(img, beam, **kwargs):
             if img.size > 0:
                 out = np.zeros(img.shape, dtype=img.dtype)
                 for index in range(img.shape[0]):
@@ -1374,7 +1377,8 @@ class DaskVaryingResolutionSpectralCube(DaskSpectralCubeMixin, VaryingResolution
         # Rechunk so that there is only one chunk in the image plane
         cube = self._map_blocks_to_cube(convfunc,
                                         additional_arrays=(beams,),
-                                        rechunk=('auto', -1, -1))
+                                        rechunk=('auto', -1, -1),
+                                        **kwargs)
 
         # Result above is a DaskVaryingResolutionSpectralCube, convert to DaskSpectralCube
         newcube = DaskSpectralCube(data=cube._data,
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -461,7 +461,8 @@ class Projection(LowerDimensionalObject, SpatialCoordMixinClass,
         if filename is not None:
             self.figure.savefig(filename)
 
-    def convolve_to(self, beam, convolve=convolution.convolve_fft):
+    def convolve_to(self, beam, convolve=convolution.convolve_fft,
+                    **kwargs):
         """
         Convolve the image to a specified beam.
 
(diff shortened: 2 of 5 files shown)


keflavich:
This change now passes `kwargs` from `convfunc` to `convolve` instead of taking the `kwargs` from `convolve_to` and passing them directly to `convolve`.  I think the behavior will end up exactly the same in both cases unless `apply_function_parallel_spatial` passes additional kwargs.  I don't think it does (or should).

@e-koch or @astrofrog, want to chime in here?  I'm not sure there's any difference, but this is a little subtle and I could be missing something.

@akleroy was this motivated by any observed behavior that was causing a problem?  You didn't mention a specific issue in #630, but if this was causing a problem, we need to add a test so we can prevent a regression.

e-koch:
I think none of the convolution kwargs were being passed (e.g., `allow_huge`, `nan_treatment`, etc). 

And the kwargs weren't being passed here but inside `apply_function_parallel_spatial`. The ordering seems to be used elsewhere so this seems fine to me: https://github.com/radio-astro-tools/spectral-cube/blob/190df4216538cd65b11ebb1fb07c261482b4a9e8/spectral_cube/spectral_cube.py#L2589

We should test this for SpectralCube, VaryingResSpectralCube, and their equivalent Dask counterparts, though.

keflavich:
Why wouldn't `allow_huge`, `nan_treatment`, etc. get passed in the current code?  `convolve` is taking `**kwargs` directly from `convolve_to`, so if you specified those, they should get passed onto `convolve`, no?

e-koch:
I thought so, but I guess it's because the function doesn't get called until it's wrapped in `apply_function_parallel_spatial`. I can reproduce the issues @akleroy was having:

```
In [3]: outcube1 = cube.convolve_to(Beam(3 * u.arcsec), nan_treatment='fill')                                      
TypeError                                 Traceback (most recent call last)
<ipython-input-3-29a0b370cc7e> in <module>
----> 1 outcube1 = cube.convolve_to(Beam(3 * u.arcsec), nan_treatment='fill')

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/utils.py in wrapper(self, *args, **kwargs)
     47                           PossiblySlowWarning
     48                          )
---> 49         return function(self, *args, **kwargs)
     50     return wrapper
     51 

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in convolve_to(self, beam, convolve, update_function, **kwargs)
   3127 
   3128         newcube = self.apply_function_parallel_spatial(convfunc,
-> 3129                                                        **kwargs).with_beam(beam)
   3130 
   3131         return newcube

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in apply_function_parallel_spatial(self, function, num_cores, verbose, use_memmap, parallel, **kwargs)
   2849                                                   num_cores=num_cores,
   2850                                                   use_memmap=use_memmap,
-> 2851                                                   **kwargs)
   2852 
   2853     def apply_function_parallel_spectral(self,

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in _apply_function_parallel_base(self, iteration_data, function, applicator, num_cores, verbose, use_memmap, parallel, memmap_dir, update_function, **kwargs)
   2767                                                               function,
   2768                                                               **kwargs)
-> 2769                                           for arg in iteration_data)
   2770             except ImportError:
   2771                 if num_cores is not None and num_cores > 1:

~/anaconda3/lib/python3.6/site-packages/joblib/parallel.py in __call__(self, iterable)
    981             # remaining jobs.
    982             self._iterating = False
--> 983             if self.dispatch_one_batch(iterator):
    984                 self._iterating = self._original_iterator is not None
    985 

~/anaconda3/lib/python3.6/site-packages/joblib/parallel.py in dispatch_one_batch(self, iterator)
    823                 return False
    824             else:
--> 825                 self._dispatch(tasks)
    826                 return True
    827 

~/anaconda3/lib/python3.6/site-packages/joblib/parallel.py in _dispatch(self, batch)
    780         with self._lock:
    781             job_idx = len(self._jobs)
--> 782             job = self._backend.apply_async(batch, callback=cb)
    783             # A job can complete so quickly than its callback is
    784             # called before we get here, causing self._jobs to

~/anaconda3/lib/python3.6/site-packages/joblib/_parallel_backends.py in apply_async(self, func, callback)
    180     def apply_async(self, func, callback=None):
    181         """Schedule a func to be run"""
--> 182         result = ImmediateResult(func)
    183         if callback:
    184             callback(result)

~/anaconda3/lib/python3.6/site-packages/joblib/_parallel_backends.py in __init__(self, batch)
    543         # Don't delay the application, to avoid keeping the input
    544         # arguments in memory
--> 545         self.results = batch()
    546 
    547     def get(self):

~/anaconda3/lib/python3.6/site-packages/joblib/parallel.py in __call__(self)
    259         with parallel_backend(self._backend):
    260             return [func(*args, **kwargs)
--> 261                     for func, args, kwargs in self.items]
    262 
    263     def __len__(self):

~/anaconda3/lib/python3.6/site-packages/joblib/parallel.py in <listcomp>(.0)
    259         with parallel_backend(self._backend):
    260             return [func(*args, **kwargs)
--> 261                     for func, args, kwargs in self.items]
    262 
    263     def __len__(self):

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/spectral_cube.py in _apply_spatial_function(arguments, outcube, function, **kwargs)
    161 
    162     if np.any(includemask):
--> 163         outcube[ii, :, :] = function(img, **kwargs)
    164     else:
    165         outcube[ii, :, :] = img

TypeError: convfunc() got an unexpected keyword argument 'nan_treatment'
```

keflavich:
OK, I see - `kwargs` is getting passed both ways.

I think this PR is a fix, so we should go ahead with it, but the test you've shown is needed too.

e-koch:
Agreed. But we should check if the fix is needed for all the `convolve_to` functions here, too. I'm happy to do so but I'll be delayed by a couple days.

keflavich:
yes, sorry, I mean we should go ahead w/expanding this PR to cover the other cases

e-koch:
Tests are passing. The failure is from me accidentally pushing a branch to upstream (deleted now).

keflavich:
Thanks @e-koch!  Thanks @akleroy!  Glad to see you're using it.

PR #638, opened 2020-06-05
EXP: Fix test suite and ensure _data is a dask array

astrofrog:
For now this is just for debugging but some things here could be merged once any issues are ironed out


keflavich:
Looks fine.  We could use some test coverage for this (it was presumably written to address a particular issue?), but I'm merging.

PR #642, opened 2020-06-15
ONLY use the Jy<->K factors when converting to K!

keflavich:
Serious bug discovered by @djeff1887.  If you convert Jy/beam to a self-equivalent unit, like mJy/beam or Jy/sr, it will convert using the Jy<->K equivalency factors, which is not at all what is intended.


keflavich:
Why is coverage dropping to zero?

keflavich:
This build is green, but github is incorrectly reporting that travis is reporting yellow (in progress).   Merging!

PR #646, opened 2020-06-20
bugfix: writeto function with overwrite does not work

1054:
Dear developers, 
The current version of spectral_cube seems could not overwrite fits cubes when calling the writeto function with overwrite=True. The issue is because in "spectral_cube/io/fits.py", inside `def write_fits_cube`, the `overwrite` argument is not passed to the called function `hdulist.writeto`. A simple fix is done within this pull request. 
Thanks for having a look.

### DIFF
--- a/spectral_cube/io/fits.py
+++ b/spectral_cube/io/fits.py
@@ -235,7 +235,7 @@ def write_fits_cube(cube, filename, overwrite=False,
             hdulist[0].header.add_history("Written by spectral_cube v{version} on "
                                         "{date}".format(version=SPECTRAL_CUBE_VERSION,
                                                         date=now))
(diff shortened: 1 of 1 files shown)


keflavich:
Thanks for this fix.  It is redundant with https://github.com/radio-astro-tools/spectral-cube/pull/640, which fixes the same issue.  I need to figure out why tests are failing before we can merge that, though.

keflavich:
Closed by #640.  Thanks @1054!  Please do contribute more like this - it's really helpful to have feedback showing when we make small mistakes.  Most of the time you'll catch them before we do, just this time, we happened to have a fix ready to go but tests were broken for another reason.

PR #651, opened 2020-07-21
Improve the error message when all beams are masked out

keflavich:
The error message now will just be "this is a bug".  Change to, "you've masked everything out"


e-koch:
@keflavich -- can this be merged now?

PR #656, opened 2020-08-15
WIP: Use common beam instead of beam averaging

e-koch:
Remove beam averaging and force operations to use the common beam. This depends on improvements to the common beam algorithm in radio-beam (https://github.com/radio-astro-tools/radio-beam/pull/81).

* [x] Remove current usage of beam averaging
* [ ] Generalize identification of bad beams to difference in areas instead of deviation from median. *requires changes in radio-beam*
* [ ] Toggle to enforce common beam operations vs. allow small deviations. With dask, would add spatial convolution operations before computing and/or introduce a "hidden" intermediate dump of a zarr file with the convolved cube (may be unfeasible, we'll see)
* [ ] Cache common beam somewhere if `goodbeam_mask` doesn't change

### DIFF
--- a/spectral_cube/base_class.py
+++ b/spectral_cube/base_class.py
@@ -541,6 +541,8 @@ class MultiBeamMixinClass(object):
                                   pa=mid_value(props['pa'])
                                  )
 
+        # Change to deviation in areas with respect to a pixel area.
+
         for prop in criteria:
             val = props[prop]
             mid = getattr(reference_beam, prop)
@@ -553,13 +555,16 @@ class MultiBeamMixinClass(object):
 
         return includemask
 
-    def average_beams(self, threshold, mask='compute', warn=False):
+    def get_common_beam(self, threshold, mask='compute', warn=False,
+                       combeam_kwargs={}):
         """
-        Average the beams.  Note that this operation only makes sense in
-        limited contexts!  Generally one would want to convolve all the beams
-        to a common shape, but this method is meant to handle the "simple" case
-        when all your beams are the same to within some small factor and can
-        therefore be arithmetically averaged.
+        Use a common beam for operations.
+
+        Many cubes will have a beam that varies by a small factor (less than a single
+        spatial pixel area) across spectral channels. In that case, this method will
+        handle avoid spatially-convolving to an exact common beam, which is an expensive
+        operation. To avoid this behaviour, a cube should be convolved to a common beam
+        prior to applying further operations.
 
         Parameters
         ----------
@@ -571,6 +576,8 @@ class MultiBeamMixinClass(object):
             and edge beams.
         warn : bool
             Warn if successful?
+        combeam_kwargs : dict
+            Additional kwargs for the common beam algorithm.
 
         Returns
         -------
@@ -601,7 +608,8 @@ class MultiBeamMixinClass(object):
 
         # use private _beams here because the public one excludes the bad beams
         # by default
-        new_beam = self._beams.average_beam(includemask=beam_mask)
+        # new_beam = self._beams.average_beam(includemask=beam_mask)
+        new_beam = self._beams.common_beam(includemask=beam_mask, **combeam_kwargs)
 
         if np.isnan(new_beam):
             raise ValueError("Beam was not finite after averaging.  "
@@ -652,7 +660,7 @@ class MultiBeamMixinClass(object):
             if need_to_handle_beams:
                 # do this check *first* so we don't do an expensive operation
                 # and crash afterward
-                avg_beam = self.average_beams(beam_threshold, warn=True)
+                com_beam = self.get_common_beam(beam_threshold, warn=True)
 
             result = function(*args, **kwargs)
 
@@ -661,14 +669,14 @@ class MultiBeamMixinClass(object):
                 return result
 
             elif need_to_handle_beams:
-                result.meta['beam'] = avg_beam
-                result._beam = avg_beam
+                result.meta['beam'] = com_beam
+                result._beam = com_beam
 
             return result
 
         return newfunc
 
-    def _check_beam_areas(self, threshold, mean_beam, mask=None):
+    def _check_beam_areas(self, threshold, common_beam, mask=None):
         """
         Check that the beam areas are the same to within some threshold
         """
@@ -691,7 +699,7 @@ class MultiBeamMixinClass(object):
         for (qtyname, qty) in (qtys.items()):
             minv = qty[mask].min()
             maxv = qty[mask].max()
-            mn = getattr(mean_beam, qtyname)
+            mn = getattr(common_beam, qtyname)
             maxdiff = (np.max(np.abs(u.Quantity((maxv-mn, minv-mn))))/mn).decompose()
 
             if isinstance(threshold, dict):


e-koch:
@astrofrog - Have you seen this `InvocationError` before? https://travis-ci.org/github/radio-astro-tools/spectral-cube/jobs/721172051 I can't reproduce it locally.

astrofrog:
I’ll try and take a look shortly!

e-koch:
@astrofrog Fixed it (somehow)! I think a file wasn't being closed when the test failed.

spectral_cube/base_class.py
```diff
@@ -554,30 +561,53 @@ def identify_bad_beams(self, threshold, reference_beam=None,
...
+                      "This is because beam averaging is not the correct operation. Please use"
+                      " `cube.set_common_beam` in its place to compute the common beam. The "
+                      "common beam can then be accessed with `cube.common_beam`.",
+                      DeprecationWarning)
+
+        self.set_common_beam(threshold, mask=mask, warn=warn)
+
+    def set_common_beam(self,
+                        threshold=None, mask='compute', warn=False,
+                        combeam_kwargs={}):
```
astrofrog, line 578: I'm not a fan of the ``combeam`` abbreviation - maybe just ``common_beam_kwargs``, or why not simply have ``**kwargs`` and pass them through to the common beam algorithm? Is there a risk of overlap? Since the main purpose of the method is to actually call the common beam algorithm, just using ``**kwargs`` would seem more natural to me?

e-koch: Agreed. I've switched to `**kwargs` here.

e-koch:
Tests on travis are giving an `InvocationError` again when one test using a dask cube fails and I think the file doesn't get closed.

Here's the failure I get locally:
```
___________________________________________________________________ test_unmasked_channels[True-data_vda] ___________________________________________________________________

filename = PosixPath('/tmp/pytest-of-eric/pytest-60/test_unmasked_channels_True_da0/vda.fits'), use_dask = True

    @pytest.mark.parametrize('filename', ['data_vda', 'data_vda_beams'],
                             indirect=['filename'])
    def test_unmasked_channels(filename, use_dask):
    
        cube, data = cube_and_raw(filename, use_dask=use_dask)
    
        # Add a mask to the cube
        mask = np.ones(cube.shape, dtype=bool)
        mask[0] = False
        cube = cube.with_mask(mask)
    
>       np.testing.assert_equal(cube.unmasked_channels, mask.any(axis=(1, 2)))

spectral_cube/tests/test_spectral_cube.py:2570: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
spectral_cube/utils.py:18: in wrapper
    self._cache[(func, args)] = func(self, *args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = DaskSpectralCube with shape=(4, 3, 2) and unit=K and chunk size (4, 3, 2):
 n_x:      2  type_x: RA---SIN  unit_x: deg...094 deg:   29.935209 deg
 n_s:      4  type_s: VOPT      unit_s: km / s  range:     -321.215 km / s:    -317.350 km / s

    @property
    @cached
    def unmasked_channels(self):
        if isinstance(self._data, da.Array):
>           return self._compute(da.any(self._mask_include, axis=(1, 2)))
E           AttributeError: 'DaskSpectralCube' object has no attribute '_mask_include'

spectral_cube/base_class.py:353: AttributeError
========================================================================== short test summary info ==========================================================================
FAILED spectral_cube/tests/test_spectral_cube.py::test_unmasked_channels[True-data_vda] - AttributeError: 'DaskSpectralCube' object has no attribute '_mask_include'

```

The non-varying resolution cube isn't getting the mask set at an earlier stage. But the test with a VRSC is fine.

e-koch:
@astrofrog @keflavich -- ready for review

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+Finding and convolving to a common resolution
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When applying spectral operations on a
+`~spectral_cube.spectral_cube.VaryingResolutionSpectralCube`, it is often
+necessary to first convolve the data to have a common beam. To retain the
+maximum amount of spatial information, the data can be convolved to the
+smallest common beam that all beams in
+`~spectral_cube.spectral_cube.VaryingResolutionSpectralCube.beam` can be
+deconvolved from.
```
keflavich, line 81: this is... pretty hard to parse.  Try to rephrase?  Maybe just end at "smallest common beam"?  Isn't it "smallest beam that all beams can be convolved to" anyway?

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+
+Finding the smallest common beam and convolution operations are described
```
keflavich, line 83:
```suggestion
Tools for finding the smallest common beam and convolution operations are described
```

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+in detail in :doc:`smoothing`.
+
+
+Small beam variations: Do I need to convolve first?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Formally, operations on multiple spectral channels require the data to have a common
+resolution. In practice, data cubes with a high-spectral resolution will have the beam
```
keflavich, line 91:
```suggestion
resolution. In practice, data cubes with a high spectral resolution will have the beam
```

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+the beams will be ignored, a warning will be printed describing this approximation, and
+the spectral operation will be computed.
+
+The beam threshold can be changed from the default 0.01 with::
+
+    >>> vrsc_cube.beam_threshold = 0.02   # doctest: +SKIP
+
+.. note::
+
+    Setting `vrsc_cube.beam_threshold = 1.0` will allow all beam variations without raising an error. Large values for the beam threshold should not be used for scientific results.
```
keflavich, line 111: I think this is not true. You can have more than 100% variation in beam size, then this will still fail.  I've had this happen when there was a 7m beam in 1 channel and real 12m beams in the others

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+
+For most spectral-line data cubes covering a single spectral line, the fine resolution
+and small frequency range will often allow the above approximation to work.
+However, if you are working with wideband data, this approximation will not hold and you
+will need to convolve to a common beam, or extract spectral slabs from the data cube.
```
keflavich, line 116:
```suggestion
will need to convolve to a common beam or extract spectral slabs from the data cube.
```

keflavich (changes requested):
some minor changes.

I just want to make sure: the default behavior is not to convolve to a common beam, is it?  I want to be sure we're not triggering massive convolution operations without warning the user, since that tends to crash computers.

e-koch:
No, no auto trigger of a whole cube convolution. An error is raised if the beams deviate too much from the common beam, similar to what we already have with the median/avg beam.

docs/beam_handling.rst
e-koch: Fixed.

docs/beam_handling.rst
e-koch: Fixed. Noted that 1.0 and larger allow for large beam variations (not "all").

docs/smoothing.rst
```diff
@@ -29,12 +29,31 @@ will be different for each slice.
...
 
-    common_beam = cube.beams.common_beam()
-    new_cube = cube.convolve_to(common_beam)
+Many spectral operations on a cube with a varying resolution will require first
+convolving to a common resolution. You may want to retain the finest spatial
+resolution possible by finding the smallest common beam that encloses all beams
+in the cube. To do this, we can use
+:meth:`~spectral_cube.VaryingResolutionSpectralCube.set_common_beam`::
+
+    cube.set_common_beam()
```
astrofrog, line 39: Is there a reason why we need to make a stateful change on the cube and set ``common_beam``? We could instead do something like:

```
common_beam = cube.find_common_beam()
cube.convolve_to(common_beam)
```

? Having it stateful means it could get out of sync if e.g. some channels are masked and so on. In general we might want to avoid stateful changes to cubes if we can?

With ``find_common_beam`` above one could even find a common beam with different settings and compare them without having to make a stateful change each time.

spectral_cube/base_class.py
```diff
@@ -507,8 +514,10 @@ def identify_bad_beams(self, threshold, reference_beam=None,
 
         Parameters
         ----------
-        threshold : float
-            Fractional threshold
+        threshold : float, optional
+            The fractional difference between beam major, minor, and pa to
```
astrofrog, line 518: maybe spell out position angle?

spectral_cube/base_class.py
```diff
@@ -688,38 +741,79 @@ def _check_beam_areas(self, threshold, mean_beam, mask=None):
...
-    def mask_out_bad_beams(self, threshold, reference_beam=None,
+            if raise_error or self.strict_beam_match:
+                raise ValueError(f"{strictmessage}{errormessage}\nConvolve to a common beam before applying any"
+                                 " spectral operation.")
+
+            else:
+                warnings.warn(errormessage)
+                warnings.warn(strictmessage)
+                warnings.warn("Convolution to a common beam will be triggered at an intermediate level."
+                              " To avoid this step, first convolve the spectral-cube to a common beam size.")
```
astrofrog, line 785: spectral-cube -> spectral cube or SpectralCube (with the hyphen is the package name?)

astrofrog (changes requested):
Looks good overall! A question below about API, in particular I think it would be cleaner to return a common beam when it is computed than store it in a stateful way on the cube.

docs/smoothing.rst
e-koch: The only issue is that the common beam operation can be slow for cubes with many (>1e3) channels. And baked into the checks for deviations is the need to check for variation from the common beam, where we had the average beam before. That's our only check (currently) for whether to ignore small beam changes for operations along the spectral axis.

Allowing a stateful change could also work if we force the beam masking to not be stateful (which I think it can be now)

docs/beam_handling.rst
```diff
@@ -53,19 +53,122 @@ Multi-beam cubes
...
+of 0.01; see above).
+However, the comparison here is to the **median** major and minor axis rather than
+the smallest common beam used above.
+This is because bad beams are identified as outliers in the set of beams.
+
+To mask the channels with bad beams, use
+`~spectral_cube.spectral_cube.VaryingResolutionSpectralCube.mask_out_bad_beams`.
+
+
+    >>> masked_vrsc_cube = vrsc_cube.mask_out_bad_beams()  # doctest: +SKIP
```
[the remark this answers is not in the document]
e-koch: `mask_out_bad_beams` was in before this PR. I could a deprecation warning here? I agree with `with_bad_beams_masked`

spectral_cube/base_class.py
e-koch: :+1:

docs/beam_handling.rst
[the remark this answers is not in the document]
astrofrog: Yes I think a deprecation warning would make sense, @keflavich do you agree?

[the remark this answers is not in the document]
keflavich: yep, :+1: to deprecating.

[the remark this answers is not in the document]
e-koch: Done!

e-koch:
* [ ] Tests for loading varying-res cubes with common beam computed (`compute_commonbeam=True`)
* [ ] Docs for loading with `compute_commonbeam=True`
* [ ] Test for `with_bad_beams_masked` (replacing state change in `identify_bad_beams`)
* [ ] Update docs for `with_bad_beams_masked`
* [ ] Test for new `convolve_to_commonbeam`
* [ ] Add docs to convolution operations for `convolve_to_commonbeam`

keflavich:
Discussion today: @e-koch will try splitting up this PR to avoid state-change related issues.

Maybe need to remove median(beams) operation

PR #661, opened 2020-09-02
WIP: Change LDOs to use masked products by default

e-koch:
This supersedes the first attempts in #540. 

Our issue stems from subclassing `u.Quantity` for all LDOs (Projection, Slice, OneDSpectrum) while trying to use the masking routines built for spectral-cube (`LDO.filled_data[:]`, `LDO._get_filled_data(np.NaN)`).

Basically `LDO.filled_data[:]` should be used by default for `LDO[slice]`, `LDO.value`, `LDO.quantity`, etc. as this matches the behaviour used for `SpectralCube`. But this interferes with how we currently have `u.Quantity` subclassed.

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -17,7 +17,7 @@ from astropy.io.registry import UnifiedReadWriteMethod
 from . import spectral_axis
 from .io.core import LowerDimensionalObjectWrite
 from .utils import SliceWarning, BeamWarning, SmoothingWarning, FITSWarning
-from .cube_utils import convert_bunit
+from . import cube_utils
 from . import wcs_utils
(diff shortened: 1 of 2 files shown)


e-koch:
@keflavich @astrofrog help would be appreciated here!

This may require disentangling `u.Quantity` as a subclass for LDOs, closer to what we use for cubes. A lot of the LDO code is already confusing/hacked together.

The current changes fix the primary issue of returning the data filled from the mask but fall apart when slicing an LDO as `__new__` doesn't get called and most attributes don't get set.

keflavich:
Questions from telecon:

 * Can we remove `Quantity` as a parent class?  What capabilities would we need to add to a base class / mixin class to minimize breakage?
 * Alternatively, can we override `.value` to use the spectral-cube masking scheme?

If you try any of these approaches, keep notes here on what breaks.

PR #662, opened 2020-09-03
Update spectral_cube.py

Jiangxuejian:
add three parameters for plot_channel_maps so that one can plot and adjust a color bar for the channel maps.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -3299,6 +3299,7 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
                           fig=None, fig_smallest_dim_inches=8, decimals=3, zoom=1,
                           textcolor=None, cmap='gray_r', tighten=False,
                           textxloc=0.5, textyloc=0.9,
+                          colorbar=False, colorbar_title='', cax=[0.93, 0.08, 0.02, 0.4],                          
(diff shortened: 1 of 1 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -3337,6 +3338,13 @@ def plot_channel_maps(self, nx, ny, channels, contourkwargs={}, output_file=None
             pointing center will be customizable.
         tighten : bool
             Call ``plt.tight_layout()`` after plotting?
+        colorbar : bool
+            plot a colorbar next the t
```
keflavich, line 3342:
```suggestion
            plot a colorbar
```

spectral_cube/spectral_cube.py
```diff
@@ -3415,6 +3423,12 @@ def plot_channel_maps(self, nx, ny, channels, contourkwargs={}, output_file=None
 
             axis_list.append(ax)
 
+        if colorbar:
+            cax = plt.axes(cax)
+            fig = plt.gcf()
```
keflavich, line 3428:
```suggestion
```

keflavich (changes requested):
Looks like a helpful addition.   

If you don't specify `vmin`/`vmax`, this colorbar will only be appropriate for the final panel, right?  We perhaps should force that `vmin`/`vmax` or `norm` are specified if the colorbar is included.

Issue #659, opened 2020-08-21
Fast global stats?


keflavich:
CASA's `ia.statistics` is pretty fast.  We should probably try to have a convenience function that does similar, e.g.:
```
                    min = cube.min()
                    max = cube.max()
                    mad = cube.mad_std()
                    std = cube.std()
                    sum = cube.sum()
                    mean = cube.mean()
```
but all at once.  Can `dask` handle doing all of this in a "single pass" - i.e., run all the stats computations in parallel with a single io read?

astrofrog:
Done in #663

PR #663, opened 2020-09-09
Added initial implementation of global 'single-pass' statistics

astrofrog:
This is a work in progress but the idea is to address #659 by adding a ``statistics`` method which finds stats over the whole cube (so no ``axis=`` keyword) using a single pass over chunks. This means a chunk is read, and all statistics are evaluated on it before moving to the next chunk. This can mean minimizing chunk I/O.

```python
In [1]: from spectral_cube import SpectralCube                                                                                                         

In [2]: cube = SpectralCube.read('l1448_13co_small.fits', use_dask=True)                                                                               

In [3]: cube.statistics()                                                                                                                              
Out[3]: 
{'min': <Quantity -0.66045946>,
 'max': <Quantity 4.0023365>,
 'sum': <Quantity 303203.88>,
 'mean': <Quantity 0.51889595>,
 'std': <Quantity 0.57547368>}
```

This is still incomplete, in particular it needs:

* [x] Changelog entry
* [x] Tests
* [x] Other 'easy' stats, e.g. count (of non-masked items) and rms
* [x] Documention
* [ ] A more robust implementation for std
* [ ] An equivalent method on the non-dask class? (not sure about this since it wouldn't be 'efficient' in the same way as for dask)

Maybe in a separate PR we can then work on adding optional 'robust' statistics (like for CASA's ``ia.statistics``) but this requires first implementing algorithms for parallel percentiles (I have ideas about this).

In the mean time, @e-koch and @keflavich, maybe you could test this out to see if the performance is reasonable (and how it compares to CASA).

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -719,11 +719,16 @@ class DaskSpectralCubeMixin:
 
         data = self._get_filled_data(fill=np.nan)
 
+        try:
+            from bottleneck import nanmin, nanmax, nansum
+        except ImportError:
+            from numpy import nanmin, nanmax, nansum
+
         def compute_stats(chunk, *args):
-            return np.array([np.nanmin(chunk),
-                             np.nanmax(chunk),
-                             np.nansum(chunk),
-                             np.nansum(chunk * chunk)])
+            return np.array([nanmin(chunk),
+                             nanmax(chunk),
+                             nansum(chunk),
+                             nansum(chunk * chunk)])
 
         results = da.map_blocks(compute_stats, data).compute()
 
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -730,7 +730,8 @@ class DaskSpectralCubeMixin:
                              nansum(chunk),
                              nansum(chunk * chunk)])
 
-        results = da.map_blocks(compute_stats, data).compute()
+        with dask.config.set(**self._scheduler_kwargs):
+            results = da.map_blocks(compute_stats, data).compute()
 
         min_values, max_values, sum_values, ssum_values = results.reshape((-1, 4)).T
 
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -735,7 +735,7 @@ class DaskSpectralCubeMixin:
                              nansum(chunk * chunk)])
 
         with dask.config.set(**self._scheduler_kwargs):
-            results = da.map_blocks(compute_stats, data).compute()
+            results = da.map_blocks(compute_stats, data.reshape(-1)).compute()
 
         count_values, min_values, max_values, sum_values, ssum_values = results.reshape((-1, 5)).T
 
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -707,6 +707,43 @@ class DaskSpectralCubeMixin:
         """
         return self._compute(da.nanargmin(self._get_filled_data(fill=np.inf), axis=axis))
 
+    @ignore_warnings
+    def statistics(self):
+        """
+        Return a dictinary of global basic statistics for the data.
+
+        This method is designed to minimize the number of times each chunk is
+        accessed. The statistics are computed for each chunk in turn before
+        being aggregated.
+        """
+
+        data = self._get_filled_data(fill=np.nan)
+
+        def compute_stats(chunk, *args):
+            return np.array([np.nanmin(chunk),
+                             np.nanmax(chunk),
+                             np.nansum(chunk),
+                             np.nansum(chunk * chunk)])
+
+        results = da.map_blocks(compute_stats, data).compute()
+
+        min_values, max_values, sum_values, ssum_values = results.reshape((-1, 4)).T
+
+        stats = {'min': min_values.min(),
+                 'max': max_values.max(),
+                 'sum': sum_values.sum()}
+
+        stats['mean'] = stats['sum'] / data.size
+
+        # FIXME: for now this uses the simple 'textbook' algorithm which is not
+        # numerically stable, so this should be replaced by a more robust approach
+        stats['std'] = ((ssum_values.sum() - stats['sum'] ** 2 / data.size) / (data.size - 1)) ** 0.5
+
+        for key in stats:
+            stats[key] = u.Quantity(stats[key], self._unit)
+
+        return stats
+
     def _map_blocks_to_cube(self, function, additional_arrays=None, fill=np.nan, rechunk=None, **kwargs):
         """
         Call dask's map_blocks, returning a new spectral cube.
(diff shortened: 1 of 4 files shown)


e-koch:
@astrofrog -- I'm getting the following error. Might be related to the block sizes, but they're all the same size (i.e. no rounding up/down # of pix) so I'm unsure where this is coming from.

```
In [12]: out = cube.statistics()                                                                                   
[########################################] | 100% Completed |  7min  1.1s
[########################################] | 100% Completed |  7min  1.2s
IndexError                                Traceback (most recent call last)
<ipython-input-12-4cf117b4bb9b> in <module>
----> 1 out = cube.statistics()

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/dask_spectral_cube.py in wrapper(self, *args, **kwargs)
     64         with warnings.catch_warnings():
     65             warnings.simplefilter('ignore')
---> 66             return function(self, *args, **kwargs)
     67 
     68     return wrapper

~/ownCloud/code_development/radio_astro_tools/spectral-cube/spectral_cube/dask_spectral_cube.py in statistics(self)
    726                              np.nansum(chunk * chunk)])
    727 
--> 728         results = da.map_blocks(compute_stats, data).compute()
    729 
    730         min_values, max_values, sum_values, ssum_values = results.reshape((-1, 4)).T

~/anaconda3/lib/python3.6/site-packages/dask/base.py in compute(self, **kwargs)
    164         dask.base.compute
    165         """
--> 166         (result,) = compute(self, traverse=False, **kwargs)
    167         return result
    168 

~/anaconda3/lib/python3.6/site-packages/dask/base.py in compute(*args, **kwargs)
    436     postcomputes = [x.__dask_postcompute__() for x in collections]
    437     results = schedule(dsk, keys, **kwargs)
--> 438     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    439 
    440 

~/anaconda3/lib/python3.6/site-packages/dask/base.py in <listcomp>(.0)
    436     postcomputes = [x.__dask_postcompute__() for x in collections]
    437     results = schedule(dsk, keys, **kwargs)
--> 438     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    439 
    440 

~/anaconda3/lib/python3.6/site-packages/dask/array/core.py in finalize(results)
    953     while isinstance(results2, (tuple, list)):
    954         if len(results2) > 1:
--> 955             return concatenate3(results)
    956         else:
    957             results2 = results2[0]

~/anaconda3/lib/python3.6/site-packages/dask/array/core.py in concatenate3(arrays)
   4313     if not ndim:
   4314         return arrays
-> 4315     chunks = chunks_from_arrays(arrays)
   4316     shape = tuple(map(sum, chunks))
   4317 

~/anaconda3/lib/python3.6/site-packages/dask/array/core.py in chunks_from_arrays(arrays)
   4093 
   4094     while isinstance(arrays, (list, tuple)):
-> 4095         result.append(tuple([shape(deepfirst(a))[dim] for a in arrays]))
   4096         arrays = arrays[0]
   4097         dim += 1

~/anaconda3/lib/python3.6/site-packages/dask/array/core.py in <listcomp>(.0)
   4093 
   4094     while isinstance(arrays, (list, tuple)):
-> 4095         result.append(tuple([shape(deepfirst(a))[dim] for a in arrays]))
   4096         arrays = arrays[0]
   4097         dim += 1

IndexError: tuple index out of range

```

e-koch:
Running more tests now, but I'm finding 7.5 min for `cube.statistics()` (at least until the above error at the end) and ~2 min for CASA's `ia.statistics()` for the same FITS cube.

astrofrog:
@e-koch - could you try again, with bottleneck installed? I've now made sure we use it if possible which should speed things up. Also could you share your cube (could do privately on slack) so that I can try and reproduce the error you are seeing?

astrofrog:
@e-koch @keflavich - for now I've made sure we use the same names of stats as for CASA. I haven't implemented any of the robust ones but presumably that can wait for a separate follow-up PR? The main missing one otherwise is flux and the coordinate-related ones - do we want to add those? (again, maybe a follow-up PR?)

astrofrog:
Not sure what's going on with the Windows build and the open file, but I'll investigate once the PR is otherwise ready

spectral_cube/tests/test_dask.py
```diff
@@ -58,3 +74,39 @@ def test_rechunk(data_adv):
     # note last element is 2 because the chunk size we asked for
     # is larger than cube - this is fine and deliberate in this test
     assert cube_new._data.chunksize == (1, 2, 2)
+
+
+def test_statistics(data_adv):
+    cube = DaskSpectralCube.read(data_adv)
+    stats = cube.statistics()
```
e-koch, line 81: @astrofrog I can reproduce (locally) the `IndexError` I commented on by forcing multiple chunks here:
```
    cube = cube.rechunk(chunks=(1, 3, 2))
```

astrofrog:
@e-koch - thanks for the suggestion for reproducing this. I think I've fixed it now, so let me know if it works for you now!

spectral_cube/dask_spectral_cube.py
```diff
@@ -707,6 +707,54 @@ def argmin(self, axis=None, **kwargs):
...
+                             nansum(chunk),
+                             nansum(chunk * chunk)])
+
+        with dask.config.set(**self._scheduler_kwargs):
+            results = da.map_blocks(compute_stats, data.reshape(-1)).compute()
+
+        count_values, min_values, max_values, sum_values, ssum_values = results.reshape((-1, 5)).T
+
+        stats = {'npts': count_values.sum(),
+                 'min': min_values.min() * self._unit,
```
e-koch, line 743: This should use `self.unit`, I think. When `self.unit` is dimensionless, `self._unit` is `None` and this throws a TypeError

e-koch:
@astrofrog Working with multiple chunks now! I've added a comment above about units (when the cube is dimensionless).

astrofrog:
Thanks! With bottleneck installed, how does the performance compare with CASA?

e-koch:
@astrofrog I'm getting a fairly large variance in run times in both CASA and spectral-cube, but the performance looks comparable!

astrofrog:
Ok so at this point we are probably I/O limited, which is good to know!

astrofrog:
Maybe we can consider merging this and improving the std calculation and adding robust stats in a follow-up.

Just one item for discussion now maybe: do we need an equivalent statistics on the non-dask class?

keflavich:
Yes and yes.  So we should merge this now, and... adding those stats for non-dask is pretty trivial, right?

astrofrog:
@keflavich - yes we can merge this. Adding it to non dask is only trivial if you are ok with whole cubes being loaded into memory!

keflavich:
mean, max, min, sum are all doable as ray/slice iterators with no change.  `std` is implementable via those iterators, but I don't think we have that implementation in hand yet.  We can do the trivial implementation first, but it would be nice to have the iterator version available too.

astrofrog:
Ok I can try and add that in a separate PR?

PR #666, opened 2020-09-22
Improve subcube WCS selection

e-koch:
`SpectralCube.subcube` allows spatial WCS coordinates to be given to slice out a subcube. However, the current code solves for the pixel coordinates of the upper and lower coordinates in each spatial dimension independently, and it uses 1D slices of WCS coords along the first pixel in each dimension. For images that cover a larger region on the sky (>~1 deg), this can lead to measurable deviations between the pixel slices for the subcube and the actual WCS limits requested.

I encountered this issue making subcubes for some of the spectral-cube tutorials. I created a subcube from one cube and used the `cube.latitude_extrema` to make an equivalent subcube from a separate cube. There ended up being a noticable ~10 arcsec (10 pixel) offset between the WCS coords in the second subcube from the first:

```
cube2_cutout = cube2.subcube(xlo=cube1_cutout.longitude_extrema[1],
                             xhi=cube1_cutout.longitude_extrema[0],
                             ylo=cube1_cutout.latitude_extrema[0],
                             yhi=cube1_cutout.latitude_extrema[1])
```

Here are the WCS coords for each spatial pixel in the two subcubes:

subcube_spatial_offset

To fix this, I've changed `SpectralCube.subcube` and `Projection.subimage` to solve for the spatial coordinate of each corner of the subcube together whenever either `x` or `y` coordinates are given in WCS units (i.e., the minimum distance between the corner and the spatial WCS coordinates for the whole grid `SpectralCube.spatial_coordinate_map`)

subcube_spatial_offset_fixed

The offset is now <1 pixel difference, which is what we expect for slicing instead of a full reprojection to match two cubes.

Because it relies on minimizing the spatial distance, this will slow down `SpectralCube.subcube` for large spatial sizes.


e-koch:
* [ ] Clean up unused code
* [ ] Add description to docs or docstring

e-koch:
@keflavich -- Here's a brief demo of where that offset was coming from.

Below is a demonstration of where the ~10 arcsec deviation in the above plots is coming from. The plot shows a pixel slice taken along the Dec axis from the `cube.spatial_coordinate_map[0]` at the centre of the map (5250 / 2):

```
dec, ra = hi_cube.spatial_coordinate_map

import matplotlib.pyplot as plt
plt.plot(dec[5250//2])
plt.xlabel("Pixel position along Dec slice")
plt.ylabel('Declination')
plt.grid(True)
plt.text(3500, 30.6598, f"Maximum deviation\n{np.ptp(dec[5250//2]).to(u.arcsec):.2f}")
```

m33_hi_BCconfig_dec_wcs_slice_variation

The ~10 arcsec deviation is the worst case in this example because the previous `cube.subcube` solver used a 1D slice taken along the 0th pixel along each direction and I was slicing out something near to the centre of the cube. So essentially the min and max in the plot.

The deviation is the sky curvature as the M33 HI mosaic I'm creating the subcube from spans 1.5 deg in dec. So it is reasonably large. In this case, the maximum deviation is ~ 1 beam element. This effect will be far more subtle for smaller images but could produce ~few pixel offsets.

By solving for each corner of the subcube together in both spatial dimensions, the subcube slicing assumes that the location of each corner can use a flat-sky approximation. This should alleviate obvious mismatches in spatial dimensions for subcubes in most cases (I think), at least where slicing is appropriate versus reprojecting onto the same WCS grid of another data set.

spectral_cube/spectral_cube.py
```diff
@@ -1820,6 +1820,10 @@ def subcube(self, xlo='min', xhi='max', ylo='min', yhi='max', zlo='min',
         """
         Extract a sub-cube spatially and spectrally.
 
+        When spatial WCS dimensions are given, the spatial coordinates of the 'lo'
+        and 'hi' corners are solved together. This minimizes WCS variations due to
+        the sky curvature when slicing from a large (>1 deg) image.
```
astrofrog, line 1825: Naive question, but if xlo/xhi/ylo/yhi are all world coordinates, why not simply call world2pix for each of the four corners?

spectral_cube/spectral_cube.py
```diff
@@ -1856,7 +1898,49 @@ def subcube(self, xlo='min', xhi='max', ylo='min', yhi='max', zlo='min',
...
+            x2 = (grids['x'] - limit_dict['x' + corn])**2
+            if hasattr(x2, 'unit'):
+                x2 = x2.value
+            y2 = (grids['y'] - limit_dict['y' + corn])**2
+            if hasattr(y2, 'unit'):
+                y2 = y2.value
+
+            dist = np.sqrt(x2 + y2)
+            # Get the min posns
+            ymin, xmin = np.unravel_index(dist.argmin(), dist.shape)
```
astrofrog, line 1934: Some of this duplicates some of the code higher up - I wonder if we can somehow factor some of it out into a helper function?

astrofrog (changes requested):
Looks good overall! My main comment below is that the logic is sufficiently non-trivial that more narrative comments would help.

spectral_cube/lower_dimensional_structures.py
```diff
@@ -567,6 +567,10 @@ def subimage(self, xlo='min', xhi='max', ylo='min', yhi='max'):
         """
         Extract a region spatially.
 
+        When spatial WCS dimensions are given, the spatial coordinates of the 'lo'
```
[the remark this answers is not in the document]
e-koch: done!

spectral_cube/lower_dimensional_structures.py
```diff
@@ -589,24 +593,102 @@ def subimage(self, xlo='min', xhi='max', ylo='min', yhi='max'):
...
+            y2 = (grids['y'] - limit_dict['y' + corn])**2
+            if hasattr(y2, 'unit'):
+                y2 = y2.value
+
+            dist = np.sqrt(x2 + y2)
+            # Get the min posns
+            ymin, xmin = np.unravel_index(dist.argmin(), dist.shape)
+
+            limit_dict['y' + corn] = ymin
+            limit_dict['x' + corn] = xmin
```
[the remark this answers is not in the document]
e-koch: Added in the new `wcs_utils.find_spatial_pixel_index`

spectral_cube/spectral_cube.py
e-koch: Yeah, that's a lot simpler! I've added that in and kept the minimum distance solving only for the case of mixed pixel/WCS inputs.

spectral_cube/spectral_cube.py
e-koch: I split it out into [wcs_utils.find_spatial_pixel_index](https://github.com/radio-astro-tools/spectral-cube/pull/666/files#diff-fc85cee04a934943754a8ab34f23bf74R486).

spectral_cube/spectral_cube.py
e-koch: And we can revert back to a 1D WCS coord slice when the input for a corner is mixed (1 pixel, 1 WCS). So the full spatial coordinate comparison I had isn't needed anymore

e-koch:
Following @astrofrog's comments, I removed the full 2D spatial distance way of finding the corners. So we now have 3 cases:
1. Everything given in pixels. Easy.
2. Both coordinates for an upper or lower coords are given in WCS. Use `WCS.world_to_array_index_values` to solve
3. We have a mix of 1 pixel and 1 WCS for a corner. In this case, we can revert back to the 1D argmin but use the given pixel value to slice out the nearest row in the direction where the WCS limit was given.

This is much better than the 2D distance argmin I had implemented because we don't need the full spatial grid of coordinates in any case.

I've added explicit tests for all 3 cases.

@astrofrog I think this is ready if tests pass. I don't get why coveralls is showing such a large decrease; the lines it's reporting as uncovered are mostly unchanged in this PR.

spectral_cube/wcs_utils.py
```diff
@@ -481,3 +481,189 @@ def diagonal_wcs_to_cdelt(mywcs):
...
+
+    When spatial WCS dimensions are given as an `~astropy.units.Quantity`,
+    the spatial coordinates of the 'lo' and 'hi' corners are solved together.
+    This minimizes WCS variations due to the sky curvature when slicing from
+    a large (>1 deg) image.
+
+
+    Parameters
+    ----------
+    cube : :class:`~SpectralCube` or spatial :clas:`~Projection`
```
keflavich, line 500:
```suggestion
    cube : :class:`~SpectralCube` or spatial :class:`~Projection`
```

spectral_cube/wcs_utils.py
```diff
@@ -481,3 +481,189 @@ def diagonal_wcs_to_cdelt(mywcs):
...
+        A spectral-cube or projection/slice with spatial dimensions.
+    [xy]lo/[xy]hi : int or :class:`~astropy.units.Quantity` or ``min``/``max``
+        The endpoints to extract.  If given as a quantity, will be
+        interpreted as World coordinates.  If given as a string or
+        int, will be interpreted as pixel coordinates.
```
keflavich, line 505:
```suggestion
        interpreted as World coordinates.  If given as a ``string`` or
        ``int``, will be interpreted as pixel coordinates.
```

spectral_cube/wcs_utils.py
```diff
@@ -481,3 +481,189 @@ def diagonal_wcs_to_cdelt(mywcs):
...
+    This minimizes WCS variations due to the sky curvature when slicing from
+    a large (>1 deg) image.
+
+
+    Parameters
+    ----------
+    cube : :class:`~SpectralCube` or spatial :clas:`~Projection`
+        A spectral-cube or projection/slice with spatial dimensions.
+    [xy]lo/[xy]hi : int or :class:`~astropy.units.Quantity` or ``min``/``max``
+        The endpoints to extract.  If given as a quantity, will be
```
keflavich, line 503:
```suggestion
        The endpoints to extract.  If given as a ``Quantity``, will be
```

spectral_cube/wcs_utils.py
```diff
@@ -481,3 +481,189 @@ def diagonal_wcs_to_cdelt(mywcs):
...
+        interpreted as World coordinates.  If given as a string or
+        int, will be interpreted as pixel coordinates.
+
+    Returns
+    -------
+    limit_dict : dict
+        Pixel coordinates of [xy]lo/[xy]hi in the given `cube`.
```
keflavich, line 510:
```suggestion
        Pixel coordinates of [xy]lo/[xy]hi in the given ``cube``.
```
presumably `cube` isn't a valid rst target, we need double-backticks?

spectral_cube/wcs_utils.py
```diff
@@ -481,3 +481,189 @@ def diagonal_wcs_to_cdelt(mywcs):
...
+
+        # (1) If both pixel units, we keep in pixel units.
+        if not any([x_hasunit, y_hasunit]):
+            continue
+
+        # (2) If both WCS units, use world_to_array_index_values
+        elif all([x_hasunit, y_hasunit]):
+
+            corn_arr = np.array([limit_dict['x'+corn].value,
+                                    limit_dict['y'+corn].value])
```
keflavich, line 596:
```suggestion
            corn_arr = np.array([limit_dict['x'+corn].value,
                                 limit_dict['y'+corn].value])
```

keflavich (changes requested):
really minor things - going to merge this soon

keflavich:
Is coverage claiming that `subcube_from_regions` is getting un-covered?

https://coveralls.io/builds/34032114/source?filename=spectral_cube%2Fspectral_cube.py#L1952

keflavich:
Green! Merging!

PR #665, opened 2020-09-22
Add operations to return WCS coordinates from argmin/argmax

e-koch:
**#656 needs to be merged before this!**

Adds `SpectralCube.argmin_world` and `SpectralCube.argmax_world` to return the WCS coordinates of the argmin/max along a dimension. I've generalized a snippet from @low-sky for this.

The main use is creating "peak velocity" maps from a cube (i.e., velocity at peak intensity):
```
peak_velocity = cube.argmax_world(axis=0)
```

* [ ] Add to docs pages
* [ ] Finish docstrings

PR #680, opened 2020-10-22
Add operations to return WCS coordinates from argmin/argmax

e-koch:
Revised version in #665 that does not depend on #656

Adds `SpectralCube.argmin_world` and `SpectralCube.argmax_world` to return the WCS coordinates of the argmin/max along a dimension. I've generalized a snippet from @low-sky for this.

The main use is creating "peak velocity" maps from a cube (i.e., velocity at peak intensity):
```
peak_velocity = cube.argmax_world(axis=0)
```

* [x] Add to docs pages
* [x] Finish docstrings

### DIFF
--- a/docs/moments.rst
+++ b/docs/moments.rst
@@ -156,3 +156,7 @@ or for modeling. The `~spectral_cube.SpectralCube.argmax_world` and
 
 Along the spectral axis, `~spectral_cube.SpectralCube.argmax_world` creates the often used
 "velocity at peak intensity," which may also be called the "peak velocity."
+
+.. note:: `cube.argmax_world` and `cube.argmin_world` are currently only defined along the spectral axis,
+as the example above shows. This is because `argmax_world` and `argmin_world` operate along the pixel axes,
+but they are not independent in WCS coordinates due to the curvature of the sky.
--- a/docs/moments.rst
+++ b/docs/moments.rst
@@ -147,7 +147,7 @@ return the pixel positions of the max/min along that axis:
     >>> argmin_map = cube.argmin(axis=0)  # doctest: +SKIP
 
 These maps are useful for identifying where signal is located within the spectral cube, however,
-it is more useful to return the WCS values of those pixels for comparisons with other data sets
+it is more useful to return the WCS values of those pixels for comparison with other data sets
 or for modeling. The `~spectral_cube.SpectralCube.argmax_world` and
 `~spectral_cube.SpectralCube.argmin_world` should be used in this case:
 
--- a/docs/moments.rst
+++ b/docs/moments.rst
@@ -158,5 +158,5 @@ Along the spectral axis, `~spectral_cube.SpectralCube.argmax_world` creates the 
 "velocity at peak intensity," which may also be called the "peak velocity."
 
 .. note:: `cube.argmax_world` and `cube.argmin_world` are currently only defined along the spectral axis,
-as the example above shows. This is because `argmax_world` and `argmin_world` operate along the pixel axes,
-but they are not independent in WCS coordinates due to the curvature of the sky.
+          as the example above shows. This is because `argmax_world` and `argmin_world` operate along the
+          pixel axes, but they are not independent in WCS coordinates due to the curvature of the sky.
--- a/docs/moments.rst
+++ b/docs/moments.rst
@@ -22,25 +22,25 @@ axis::
           moment is computed as the variance. For the actual formulas used for
           the moments, please see :class:`~spectral_cube.SpectralCube.moment`.
           For linewidth maps, see the `Linewidth maps`_ section.
-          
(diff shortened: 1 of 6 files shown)


keflavich:
Where do we stand on this?  You need review?  Rebase?

e-koch:
Rebase. Will do by the end of the day

e-koch:
@astrofrog @keflavich This is ready!

astrofrog (commented):
I like this but I wonder if we should return all world coordinates - there are cases where the RA/Dec may not line up with the pixel axes, for example they may be rotated by 45 deg. So then if you called argmax_world(axis=1) it's not clear if what would be returned is RA or Dec. Picking the world coordinate with the same index as the pixel axis only make sense in the cases where pixel and world axes are aligned?

keflavich (approved):

keflavich:
@astrofrog Is this presently going to return incorrect results in the case of a rotated WCS?  If so, we should special-case this for aligned axes only (we have that ability in generalized WCS, right?), and punt the general case for now.  I'm not sure how much use there is for the general case anyway.

spectral_cube/cube_utils.py
```diff
@@ -486,3 +487,51 @@ def convert_bunit(bunit):
...
+
+    Returns
+    -------
+    out : astropy.units.Quantity
+        2D array of WCS coordinates.
+    '''
+
+    if is_proj_plane_distorted(cube.wcs):
+        raise WCSCelestialError("world_take_along_axis requires the celestial axes"
+                                " to be aligned along image axes.")
```
keflavich, line 517: From Tom, should use:
https://docs.astropy.org/en/stable/api/astropy.wcs.wcsapi.BaseLowLevelWCS.html#astropy.wcs.wcsapi.BaseLowLevelWCS.axis_correlation_matrix

e-koch:
@astrofrog @keflavich  Here's the check I added for correlated image -> WCS axes: https://github.com/radio-astro-tools/spectral-cube/pull/680/files#diff-5efbc7ba09eb93125b163d09f8107e93a24aac38ed8791ef0b17b5ffb2fbea18R486

My understanding of `axis_correlation_matrix` is that the celestial axes will _always_ be labeled as correlated, so `argmin_world`/`argmax_world` will only work for the spectral axis without.

I'm wondering if the changes in this PR should be restricted to only the spectral axis. I'm now not sure that any other case is well defined

spectral_cube/cube_utils.py
e-koch: Fixed.

e-koch:
Current failures will be fixed in #688

e-koch:
@astrofrog Can you have a quick look through the new changes? I've added checks for correlated pixel -> WCS axes.

This otherwise ready to go, IMO

docs/moments.rst
```diff
@@ -112,14 +112,51 @@ with either of these two commands::
...
+    >>> min_map = cube.min(axis=0)  # doctest: +SKIP
+
+The `~spectral_cube.SpectralCube.argmax` and `~spectral_cube.SpectralCube.argmin` operations will
+return the pixel positions of the max/min along that axis:
+
+    >>> argmax_map = cube.argmax(axis=0)  # doctest: +SKIP
+    >>> argmin_map = cube.argmin(axis=0)  # doctest: +SKIP
+
+These maps are useful for identifying where signal is located within the spectral cube, however,
+it is more useful to return the WCS values of those pixels for comparisons with other data sets
```
keflavich, line 150:
```suggestion
it is more useful to return the WCS values of those pixels for comparison with other data sets
```

spectral_cube/cube_utils.py
```diff
@@ -486,3 +487,51 @@ def convert_bunit(bunit):
             unit = None
 
     return unit
+
+
+def world_take_along_axis(cube, position_plane, axis):
+    '''
+    Apply a 2D plane of pixel positions along collapsed along a third axis
+    to the equivalent WCS coordinates. For example, this will convert `argmax`
```
keflavich, line 495:
```suggestion
    Convert a 2D plane of pixel positions to the equivalent WCS coordinates.
    For example, this will convert `argmax`
```
The pixel positions can be obtained in any way, right?  `argmax` is a good example, but this code will work on any plane of pixel positions

keflavich (approved):
lgtm, though I have suggestions on the docstring.

also, there's a lot of duplicated code between argmax/argmin.  Could we combine those into one with `method='max'`/`method='min'` and just pick between them, then have `argmin_world` just return `argminmax_world(method='max', **kwargs)` ?

e-koch:
@keflavich -- Thanks for the suggestion. I've added `_argminmax_world` with the common code.

I think this is ready to merge if these last tests pass.

PR #682, opened 2020-10-27
Improve performance of CASA reading by combining chunks on-the-fly

astrofrog:
For large cubes, the CASA dask reader becomes inefficient because a ~40Gb cube could end up having 10^6 chunks, so there is a huge overhead in dask for handling so many chunks/blocks which leads to much poorer performance compared to e.g. ``exportfits``. Ideally we should not have more than a few thousand chunks/blocks.

This PR is highly experimental - what it does is that it will combine neighboring chunks on disk in CASA files on-the-fly, so that the dask chunks are basically a multiple of CASA chunks. With this, I can get very close to ``exportfits`` in performance, and with the ``threading`` scheduler I can get better performance when exporting to FITS.

At this point it'd be great to gather some initial performance results on different systems and with different files!

**Notes:**

* The target dask chunksize can be set with:

```
cube = SpectralCube.read(..., target_chunksize=10000000)
```

That value is the maximum number of elements in a dask chunk - the largest chunk that is smaller or equal to this will be chosen.
* For now this does  not correctly reorder the bits for the masks, so masks will be wrong! I have been testing write performance without masks so this wasn't important yet and the masks are a small fraction of the runtime.
* For now this adds a C extension to do the re-ordering for maximum possible performance improvement. We should discuss once we know the performance results whether we are ok with that, whether we want to distribute that (and maybe other CASA I/O things) in a separate package, or if we want a pure-Python solution (possible but not as good performance)
* I think the C code could be simplified quite a bit - at the moment it can actually deal with arbitrarily stored Numpy arrays but in fact we know the data are contiguous

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ docs/api
 # Other generated stuff
 */version.py
 */cython_version.py
+.coverage*
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
 requires = ["setuptools",
             "setuptools_scm",
+            "oldest-supported-numpy",
             "wheel"]
 build-backend = 'setuptools.build_meta'
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,11 @@
 
 import os
 import sys
+
+import numpy
+
 from setuptools import setup
+from setuptools.extension import Extension
 
 TEST_HELP = """
 Note: running tests is no longer done using 'python setup.py test'. Instead
@@ -48,4 +52,7 @@ if 'build_docs' in sys.argv or 'build_sphinx' in sys.argv:
     print(DOCS_HELP)
     sys.exit(1)
 
-setup(use_scm_version={'write_to': os.path.join('spectral_cube', 'version.py')})
+setup(use_scm_version={'write_to': os.path.join('spectral_cube', 'version.py')},
+      ext_modules=[Extension("spectral_cube.io._casa_chunking",
+                             [os.path.join('spectral_cube', 'io', '_casa_chunking.c')],
+                             include_dirs=[numpy.get_include()])])
--- a/spectral_cube/io/_casa_chunking.c
+++ b/spectral_cube/io/_casa_chunking.c
@@ -18,7 +18,6 @@ static PyMethodDef module_methods[] = {
 
 /* This is the function that is called on import. */
 
-#if PY_MAJOR_VERSION >= 3
 #define MOD_ERROR_VAL NULL
 #define MOD_SUCCESS_VAL(val) val
 #define MOD_INIT(name) PyMODINIT_FUNC PyInit_##name(void)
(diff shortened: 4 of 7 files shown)


e-koch:
@astrofrog -- I checked this out and the performance improvement is substantial on a large (28 GB) CASA cube!

There is a problem with what I think is the dimension ordering, maybe in the mask? Here's the cube I had loaded in:
```
DaskVaryingResolutionSpectralCube with shape=(272, 5250, 5250) and unit=Jy / beam and chunk size (4, 750, 5250):
 n_x:   5250  type_x: RA---SIN  unit_x: deg    range:    22.608246 deg:   24.316279 deg
 n_y:   5250  type_y: DEC--SIN  unit_y: deg    range:    29.928025 deg:   31.386080 deg
 n_s:    272  type_s: FREQ      unit_s: Hz     range: 1420644196.644 Hz:1421967402.840 Hz
```
So small in the spectral dim but stretching one of the full spatial dimensions. The issue is when loading in a single channel, there are strips of NaNs across the map:
daskspectralcube_casa_OTF_chunking_orderingissue

The position of the strips changes along the spectral axis which is why I think the mask dim ordering might be the problem

astrofrog:
Yes the mask stuff doesn't work yet, working on it :)

e-koch:
Ah, I missed that above! Got it!

e-koch:
@astrofrog There's a lingering offset in the mask that is mostly along one dimension. Here's with the same spectral channel from above:

daskspectralcube_casa_OTF_chunking_orderingissue_10292020

The chunk sizes are (4, 525, 5250) for a (272, 5250, 5250) cube. It looks like the shift in the mask is along the dimension that has only 1 chunk.

astrofrog:
@e-koch - could you try again? I think I have fixed the mask issues.

e-koch:
@astrofrog The big offset is fixed! But there are still (i) strip-like patterns in the mask at the edge of the observed field and (ii) a near regular pattern of small "gaps" in the first channel of each spectral chunk.

Here's the whole field (colormap changed to see the pattern of gaps):

daskspectralcube_casa_OTF_chunking_orderingissue_10302020

Zoom in of (i) strip-like edges:

daskspectralcube_casa_OTF_chunking_orderingissue_10302020_edgezoom

There's a blank offset of zeros that you can see in some areas. It seems to be there along the whole map so isn't just an offset.

Zoom within the map of (ii) the pattern of small gaps:

daskspectralcube_casa_OTF_chunking_orderingissue_10302020_middlezoom

This occurs in in the first channel of each spectral chunk (chunks here are (4, 525, 5250)).

Most of these are 1x4 pixel masked regions. The pattern isn't quite perfect in some regions but is close to a grid of 106 pixels in the x-direction and 18 pixel in the y-direction. Neither are integer multiples of the chunk size of (525, 5250) in the spatial axes.

astrofrog:
Interesting! Could you share the cube?

astrofrog:
Just to check, did you recompile the extension? (e.g running pip) - the big offset was fixed in python code so I'm wondering if you were using pip editable mode and just checked out the latest commit without rerunning pip. Or could you try installing from a clean checkout?

e-koch:
You're right! I didn't recompile the extension. After a clean install, all those issues are fixed:

daskspectralcube_casa_OTF_chunking_orderingissue_10302020_fixed

astrofrog:
I am going to close this in favor of https://github.com/radio-astro-tools/spectral-cube/pull/684

PR #684, opened 2020-11-08
Use casa-formats-io package for CASA I/O

astrofrog:
For now this relies on the developer version but we should release casa-formats-io and update this PR before merging.

This indirectly introduces the significant performance improvements from #682 (since these are included in casa-formats-io)


astrofrog:
Note that unrelated to the changes here the CASA + dev dependencies job started failing because astropy dev now requires Python 3.7, so I've commented out that build.

astrofrog:
@e-koch - could you double check the use case you had in https://github.com/radio-astro-tools/spectral-cube/pull/682 to make sure the performance gains are still there and there are no issues with the mask?

keflavich (approved):

e-koch (approved):
Looks good! I'm getting similar performance between #682 and `casa-formats-io`. And no mask issues!

PR #686, opened 2020-11-10
Fix coverage reporting to not include tests

astrofrog:
The codecov coverage report currently (but should not) includes the tests, so this fixes it by moving the coverage configuration to setup.cfg in line with APE 17 recommendations.


e-koch (approved):
Looks good to me. I _think_ the codecov changes are all the test files that I see.

PR #688, opened 2020-11-19
Fix failed beam deconvolution with exception BeamError

e-koch:
I inadvertently caused a failure where from new changes in radio-beam from https://github.com/radio-astro-tools/radio-beam/pull/87.

There, I changed a failed deconvolution from `ValueError` to the more appropriate (and already existing) exception `BeamError`. However, we test for that exception in one of the beam convolution tests here.


e-koch:
@astrofrog @keflavich This is a one line change but will be needed for any tests using the dev version of radio-beam (so I think all of the tests?)

Issue #689, opened 2020-11-25
Unable to remove text using plot_channel_maps


drtobybrown:
From the [docs](https://spectral-cube.readthedocs.io/en/latest/api/spectral_cube.SpectralCube.html#spectral_cube.SpectralCube):
> plot_channel_maps(...)
>      **textcolor**: None or str
>          _Color of the label text to overlay. If None, will be determined automatically. If 'notext', no text will be added._

However in the following example this gives the error `ValueError: Invalid RGBA argument: 'notext'`

Example:

```
from astropy.utils import data
from spectral_cube import SpectralCube

fn = data.get_pkg_data_filename('tests/data/example_cube.fits', 'spectral_cube')
cube = SpectralCube.read(fn)

cube.plot_channel_maps(2, 2, np.arange(1,5), textcolor='notext')
```

spectral-cube version 0.5.0

keflavich:
looks like this line:
https://github.com/radio-astro-tools/spectral-cube/blob/71fd8e2c7251b46e2738af6fa3c0b78fd242260c/spectral_cube/spectral_cube.py#L3421
should have been in the `if textcolor != 'notext'` indented block

drtobybrown:
Yep, I see. 

I also notice that `contourcolors` kwarg is no longer in the function (incorporated into `contourkwargs={}`?) but remains in the docs and comments.

keflavich:
We'd welcome a PR to fix these!

drtobybrown:
done.

keflavich:
closed by #690

PR #690, opened 2020-11-25
Fix bug causing plot_channel_maps(..) to break when textcolor='notext' set

drtobybrown:
Fixes issue #689

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -3320,7 +3320,7 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
             List of channels to show
         cmap : str
             The name of a colormap to use for the ``imshow`` colors
-        contourkwargs : list
+        contourkwargs : dict
             Keyword arguments passed to ``contour``
         textcolor : None or str
             Color of the label text to overlay.  If ``None``, will be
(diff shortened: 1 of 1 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -3409,6 +3409,8 @@ def plot_channel_maps(self, nx, ny, channels, contourkwargs={}, output_file=None
                         textcolor = 'w'
                     else:
                         textcolor = 'k'
+    
+                ax.tick_params(direction='in', color=textcolor)
```
keflavich, line 3413:
```suggestion
                ax.tick_params(color=textcolor)
```

spectral_cube/spectral_cube.py
```diff
@@ -3418,8 +3420,6 @@ def plot_channel_maps(self, nx, ny, channels, contourkwargs={}, output_file=None
                 ax.coords[0].set_ticklabel_position('')
                 ax.coords[1].set_ticklabel_position('')
 
-            ax.tick_params(direction='in', color=textcolor)
```
keflavich, line 3421: suggestion:
```
            ax.tick_params(direction='in')
```

keflavich (changes requested):
small change: I think the ticks pointing in should happen either way.

But, otherwise, this looks good

e-koch:
I think all the test failures here are a carryover from #688

keflavich:
I'm merging this from the command line with one additional small tweak; unfortunately I can't push to this branch so I'm kinda crossing my fingers that I didn't break the build by removing whitespace and re-adding the ticks-in

keflavich:
This has been merged despite github's complaints.

PR #698, opened 2021-02-02
Update install script w/subproc

keflavich:
Our install instructions use the deprecated pip.main and shouldn't.

https://docs.astropy.org/en/latest/install.html#installing-astropy-into-casa is the "right way" to hackinstall stuff into old casa.


keflavich:
v0.4.4 is the last version compatible w/py2.7

e-koch:
Works locally for me. But I think because of the py2.7 version, we should list a separate command for CASA 5 versions pinning to the 0.4.4: `spectral-cube==0.4.4`.

e-koch (approved):

PR #705, opened 2021-03-04
Allow target_chunksize to be passed to reader when loading in CASA datasets

astrofrog:
In some cases the target_chunksize can end up being too large so this gives the option to customize it. The test will fail for now, will try and fix later today.


keflavich (approved):
lgtm.  What are the no-change changes to some files doing?

Also... can we give a specific recommendation in this document about when/where you might choose a specific chunk size?  I really don't know if I should be using 1e6, 1e7, 1e9, or what right now.  I don't have a sense, from these docs, what a chunk shape will end up being if I set a target_chunksize (for a spectral-dimension chunk)

astrofrog:
@keflavich - the 0 line changes are old data files I'm removing (now in casa-formats-io). I'll try and provide some guidance about the chunk size in the docs.

docs/dask.rst
```diff
@@ -92,6 +93,29 @@ installed.
 This can also be beneficial if you are using multiprocessing or multithreading to carry out calculations,
 because zarr works nicely with disk access from different threads and processes.
 
+Reading in CASA data and default chunk size
+-------------------------------------------
+
+CASA datasets are typically stored on disk with very small chunks - if we mapped these directly to
+dask array chunks, this would be very inefficient as the dask tree would then contain in some cases
```
e-koch, line 100:
```suggestion
dask array chunks, this would be very inefficient as the [dask task graph](https://docs.dask.org/en/latest/graphs.html) would then contain in some cases
```
for consistent naming with the dask docs + easy reference

spectral_cube/io/casa_image.py
```diff
@@ -57,14 +58,14 @@ def load_casa_image(filename, skipdata=False, memmap=True,
...
-        data = image_to_dask(filename, memmap=memmap)
+        data = image_to_dask(filename, memmap=memmap, target_chunksize=target_chunksize)
 
     # CASA stores validity of data as a mask
     if skipvalid:
         valid = None
     else:
         try:
-            valid = image_to_dask(filename, memmap=memmap, mask=True)
+            valid = image_to_dask(filename, memmap=memmap, mask=True, target_chunksize=target_chunksize)
```
e-koch, line 68: Noting for the future: I think eventually we should expand this to allow a memory limit (e.g., 8 * u.Mb) or exact array chunk shape.

e-koch (approved):
Looks good. I just added a small docs comment.

astrofrog:
@keflavich @e-koch - I've edited the docs, can you check to see if this makes sense now?

e-koch:
Looks good! The explanation for rechunking along the spectral dimension is clear.

keflavich (approved):

PR #711, opened 2021-05-04
Avoid data copy when calling sigma_clip

astrofrog:
This is a small optimization but every little helps! (I think it will only really make a difference with Astropy v4.3 as there was a copy-related bug in previous versions). I also opened https://github.com/astropy/astropy/pull/11692 which improves the multi-threading case.

In any case here are some interesting timings on dedicated hardware with a 12Gb cube and with this latest version:

```python
import dask
cube = SpectralCube.read('BrickMaser_271_spw49.image')
sclipped = cube.sigma_clip_spectrally(threshold=1.8, maxiters=None,
                                      save_to_tmp_dir='temp.zarr')
```

runs in **4min12s**. With:

```
cube.use_dask_scheduler('threads')
```

This goes down to **54s**! Now going back to serial mode with MAD:

```
sclipped = cube.sigma_clip_spectrally(threshold=1.8, maxiters=None,
                                      save_to_tmp_dir='temp.zarr',
                                      stdfunc='mad_std')
```

this takes **4min4s** so basically the same as with regular standard deviation. In comparison if I use the old astropy sigma clipping code with the ``mad_std`` function specified manually:

```
def nan_mad_std(data, axis=None):
    return mad_std(data, axis=axis, ignore_nan=True)

sclipped = cube.sigma_clip_spectrally(threshold=1.8, maxiters=None,
                                      save_to_tmp_dir='temp.zarr',
                                      stdfunc=nan_mad_std)
```

This takes **21min41s**, so in this particular case the new mad_std implementation is 5x faster.

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -809,7 +809,7 @@ class DaskSpectralCubeMixin:
 
         def spectral_sigma_clip(array):
             return stats.sigma_clip(array, sigma=threshold, axis=0,
-                                    masked=False, copy=False, **kwargs)
+                                    masked=False, copy=True, **kwargs)
 
         return self.apply_function_parallel_spectral(spectral_sigma_clip,
                                                      accepts_chunks=True)
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -808,8 +808,8 @@ class DaskSpectralCubeMixin:
         """
 
         def spectral_sigma_clip(array):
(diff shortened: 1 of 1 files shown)


keflavich:
This is great!

Are there any possible consequences to `copy=False`, though?  Is that going to modify the array inplace?

astrofrog:
It's not clear since every time you access a chunk it might return a new copy of the data, though with caching that may not be true. But let's keep the copy for now as I don't have time to investigate.

astrofrog:
I've added the copy back for now but kept the ``masked=False`` improvement

keflavich (approved):

PR #714, opened 2021-05-13
Expose more options for dask's map_blocks

e-koch:
Adds two things:
* Makes returning a new cube optional for `apply_function_parallel_spectral`
* Retains passing `block_info` for functions that need to be aware of a chunk's location in the whole cube.

We need both of these for `apply_function_parallel_spectral` to be used for spectral fitting a whole cube: https://github.com/radio-astro-tools/tutorials/pull/12

TODO:
* [x] Test non-cube return
* [x] Test passing block_info to a function

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -564,11 +564,24 @@ class DaskSpectralCubeMixin:
             Passed to ``function``
         """
 
-        def wrapper(data, block_info=None, **kwargs):
-            if data.size > 0:
-                return function(data, block_info=block_info, **kwargs)
-            else:
-                return data
+        # NOTE: `block_info` should always be available for `dask.array.map_blocks` to pass to
+        # Because we use this wrapper, this always should be an available kwarg, then we check
+        # if that kwarg should be passed to `function`
+        _has_blockinfo = 'block_info' in inspect.signature(function).parameters
+
+        # if/else to avoid an if/else in every single wrapper call.
+        if _has_blockinfo:
+            def wrapper(data, block_info=None, **kwargs):
+                if data.size > 0:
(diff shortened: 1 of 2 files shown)


spectral_cube/dask_spectral_cube.py
```diff
@@ -769,19 +805,23 @@ def _map_blocks_to_cube(self, function, additional_arrays=None, fill=np.nan, rec
             data = data.rechunk(rechunk)
 
         if additional_arrays is None:
-            newdata = data.map_blocks(function, dtype=data.dtype, **kwargs)
+            # newdata = data.map_blocks(function, dtype=data.dtype, **kwargs)
```
keflavich, line 808: probably just remove this?

keflavich (changes requested):

spectral_cube/dask_spectral_cube.py
```diff
@@ -553,31 +554,64 @@ def apply_function_parallel_spectral(self,
...
+                if data.size > 0:
+                    return function(data, block_info=block_info, **kwargs)
+                else:
+                    return data
+        else:
+            def wrapper(data, **kwargs):
+                if data.size > 0:
+                    return function(data, **kwargs)
+                else:
+                    return data
```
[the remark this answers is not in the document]
e-koch: Adding `block_info` to the kwargs isn't working, but I agree this looks simpler. `da.map_blocks` gives `block_info=None` if it isn't a defined keyword in the function. dask is using `inspect.signature(function).parameters` to find if `block_info` is included, so I suspect this is a limitation of `inspect.signature`.

spectral_cube/dask_spectral_cube.py
e-koch: Yup! thanks for catching

Issue #696, opened 2020-12-22
allow_huge=False in convolve_fft


foxmouldy:
I am trying to convolve a cube to a new beam: 

`bmaj_new = 12.`
`new_beam = radio_beam.Beam(major=bmaj_new*u.arcsec, minor=bmaj_new*u.arcsec, pa=0*u.deg)`
`cube_beam_smoothed = cube.convolve_to(new_beam)`

The cube is 4k x 4k pixels and 16 channels. 

I use `cube.allow_huge_operations = True`, but this isn't passed to convolve_fft, thus resulting in the following error:

`/anaconda3/lib/python3.6/site-packages/astropy/convolution/convolve.py in convolve_fft(array, kernel, boundary, fill_value, nan_treatment, normalize_kernel, normalization_zero_tol, preserve_nan, mask, crop, return_fft, fft_pad, psf_pad, min_wt, allow_huge, fftn, ifftn, complex_dtype)
    686         raise ValueError("Size Error: Arrays will be {}.  Use "
    687                          "allow_huge=True to override this exception."
--> 688                          .format(human_file_size(array_size_C)))
    689 
    690     # For future reference, this can be used to predict "almost exactly"
ValueError: Size Error: Arrays will be 1.0G.  Use allow_huge=True to override this exception
`

Solution: pass the value of `allow_huge_operations` in `cube.convolve_to` to `convolve_fft`

keflavich:
Good catch, thanks!  Would you like to submit your proposed fix as a pull request?

e-koch:
This is a good solution!

In case anyone else encounters this error before the fix is added, `cube.convolve_to(new_beam, allow_huge=True)` can also be used.

e-koch:
Fixed in #725

PR #725, opened 2021-06-15
Pass `self.allow_huge_operations` to convolve_fft`

keflavich:
Solves issue #696 

`astropy.convolution.convolve` doesn't accept `allow_huge`, so we have to check which function we're calling before deciding what to pass.

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1395,6 +1395,9 @@ class DaskSpectralCube(DaskSpectralCubeMixin, SpectralCube):
 
         # See #631: kwargs get passed within self.apply_function_parallel_spatial
         def convfunc(img, **kwargs):
+            if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
+                kwargs['allow_huge'] = self.allow_huge_operations
+
             return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape) * beam_ratio_factor
 
         return self.apply_function_parallel_spatial(convfunc,
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1395,10 +1395,10 @@ class DaskSpectralCube(DaskSpectralCubeMixin, SpectralCube):
 
         # See #631: kwargs get passed within self.apply_function_parallel_spatial
         def convfunc(img, **kwargs):
-            if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
-                kwargs['allow_huge'] = self.allow_huge_operations
-
             return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape) * beam_ratio_factor
+
+        if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
+            kwargs['allow_huge'] = self.allow_huge_operations
 
         return self.apply_function_parallel_spatial(convfunc,
                                                     accepts_chunks=True,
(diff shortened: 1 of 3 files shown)


e-koch (changes requested):
Looks good. It's unclear to me where the test failures are coming from, though. I ran locally and it seemed to be fine.

The only change is to take the `if` statements out of `convfunc`. It should be fine outside of the function and avoids an extra if call in the wrapper.

keflavich:
I don't think we can take the `if` statement out of the `convfunc` because of the possibility of `allow_huge` being passed via `apply_function_parallel_*` as noted in https://github.com/radio-astro-tools/spectral-cube/pull/631

e-koch:
Good point! That makes sense.

e-koch (dismissed):

keflavich:
I definitely wouldn't have caught that except for the comment one of us left a while back.

But I would like to get tests passing before merging.

e-koch:
Issue is in `test_beams_convolution`. For some reason, the fixture isn't getting read in or something? See line 64991 in the build log.

e-koch:
Issue is in `test_beams_convolution`. For some reason, the fixture isn't getting read in or something? See line 64991 in the build log.

keflavich:
Let's try to figure this out on next week's telecon

keflavich:
I'm skeptical of this build failure.  It seems to be saying that just passing along some kwargs here is making all the open-files tests fail.  No way.

e-koch:
Somehow it's a file being left open. Removing `--open-files` from the pytest call fixes things. I suspect the memory map file, though it's not a problem elsewhere?

e-koch:
Most of these failure are new, though they all happen at the same place. Before only `py39-novis-cov` failed in this way.

e-koch:
I'm still unsure _what_ the failure is caused by but it's triggered by accessing `self.allow_huge_operations` within the convolution wrapper. For example, directly setting a boolean works in my local tests:
```
        def convfunc(img, **kwargs):
            if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
                kwargs['allow_huge'] = True
            return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape) * beam_ratio_factor
```
So this must be some weird interaction of accessing `self` from within this wrapper, which gets passed to `SpectralCube.apply_function_parallel_spatial`.

I'm fairly sure we can just move the if statement outside of the wrapper to avoid this (?)

e-koch:
@keflavich  I tested this locally on a >1 GB cube. ``allow_huge`` was passed.

It would be great to test this but we only trigger the `allow_huge` flag for >1 GB sizes in convolve_fft.

keflavich:
We could make the threshold configurable (is it already configurable?) and set the threshold lower for tests

keflavich:
Oh, wow, I bet the `convfunc` is getting pickled (or in some way stored), which is putting `self` (which is the cube) into memory.  That's... not an interaction I would've predicted.

keflavich:
Should we try to add another test, or just merge this?  I'll leave it to you

e-koch:
I don't see an easy way to trigger the `allow_huge` flag in `convolve_fft`. Neither threshold is configurable :`convolve_fft` is 1 GB fixed, same with the `is_huge` check here https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/cube_utils.py#L257).

For now, I've added a bool flag to turn enable `allow_huge_operations` in a couple of the convolution tests. I'll check our list of issues and add a note to return to this when the threshold is configurable.

One of us can merge once the tests pass.

keflavich:
yeah, we could make https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/cube_utils.py#L255 configurable for testing at least.  But we can do that somewhere else.

Issue #729, opened 2021-06-24
Automatically determine acceptable (or ideally, optimal) chunk size


keflavich:
There are several issues that result from poor chunking choices:

 * https://github.com/radio-astro-tools/spectral-cube/issues/728
 * https://github.com/radio-astro-tools/spectral-cube/issues/697

Some of this results from us trying to force good chunking choices on spectral-based or spatial-based operations, but when you chain spectral onto spatial operations, this becomes non-optimal.

Can we use https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking in addition to some guidance on where chunks should be split to optimize chunking?

PR #730, opened 2021-06-24
use 'auto' instead of 1 for spectral rechunk

keflavich:
Minor correction to partly address #729

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1278,8 +1278,8 @@ class DaskSpectralCubeMixin:
             else:
                 #return np.interp(args[0], args[1], y[:, 0, 0],
                 #                 left=fill_value, right=fill_value).reshape(chunkshape)
-                interp = scipy.interpolate.interp1d(args[0], y, fill_value=fill_value)
-                return interp(args[1])
+                interp = scipy.interpolate.interp1d(args[1], y, fill_value=fill_value)
+                return interp(args[0])
 
         newcube = cubedata.map_blocks(interp_wrapper,
                                       args=(spectral_grid.value, inaxis.value),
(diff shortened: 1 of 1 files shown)


keflavich:
this apparently breaks!

keflavich:
This shows that apparently `spectral_interpolate` is changing the cube size in the _spatial_ dimensions.  ?!?!?
image

keflavich:
I switched from `np.interp` to `scipy.interpolate.interp1d` because the latter allows for straightforward broadcasting, while the former does not - in our non-dask version, we have a loop over valid spectral pixels.  This seems to work.

spectral_cube/dask_spectral_cube.py
```diff
@@ -1265,21 +1266,26 @@ def spectral_interpolate(self, spectral_grid,
...
             else:
-                return np.interp(args[0], args[1], y[:, 0, 0],
-                                 left=fill_value, right=fill_value).reshape((-1, 1, 1))
-
-        if reverse_in:
-            cubedata = cubedata[::-1, :, :]
-
-        cubedata = cubedata.rechunk((-1, 1, 1))
+                #return np.interp(args[0], args[1], y[:, 0, 0],
+                #                 left=fill_value, right=fill_value).reshape(chunkshape)
```
e-koch, line 1280: Clean up here.

e-koch:
```suggestion

```

e-koch (approved):
Looks good, just a small clean-up.

I don't think the tests need to change from the existing ones, since they serve the same purpose.

The dask and non-dask versions (so scipy vs numpy now) are equivalent from a couple local tests I ran.

Note: In a separate PR, we should make scipy a dependency. The non-dask spectral interpolate could be switched over just to match here, but there's currently no functional reason to do so.

Issue #737, opened 2021-08-21
Beam dimensions incorrectly assumed to be in arcsec for VaryingResolutionSpectralCube


drtobybrown:
When reading a cube with varying spectral resolution where BMAJ, BMIN header keyword values are in degrees (which I'd thought was standard) spectral-cube assumes they're in arcsec.

I believe the assumption is at line 3654 in spectral_cube.py

```
# CASA beam tables are in arcsec, and that's what we support
beams = Beams(major=u.Quantity(beam_data_table['BMAJ'], u.arcsec),
              minor=u.Quantity(beam_data_table['BMIN'], u.arcsec),
              pa=u.Quantity(beam_data_table['BPA'], u.deg),
              meta=[{key: row[key] for key in beam_data_table.names
                     if key not in ('BMAJ','BPA', 'BMIN')}
                    for row in beam_data_table],
             )
```

I'm not familiar enough with the code to find where the `beam_data_table` (or `beam_table`) are parsed from the headers. Beam tables in arcsec seems ok, but the header values are not being converted from degrees.

If it helps, cube I'm using is [here (updated link)](https://drive.google.com/file/d/1PqW60Z_bLDK5RRLhnfPFIvQHh1UxzDR_/view?usp=sharing)

keflavich:
We'll have to look more closely at that cube to see what the problem is.  That `__init__` code is used for FITS reading, but not for CASA `.image` reading.  I've also verified that CASA generally uses arcseconds, not degrees, in the BMIN/BMAJ columns.  Note that the BMIN/BMAJ columns are _not_ equivalent to the BMIN/BMAJ header keywords, which _are_ in degrees by default.

for CASA images, https://github.com/radio-astro-tools/spectral-cube/blob/master/spectral_cube/io/casa_image.py#L101-L126

```python
    if 'major' in beam_:
        beam = Beam(major=u.Quantity(beam_['major']['value'], unit=beam_['major']['unit']),
                    minor=u.Quantity(beam_['minor']['value'], unit=beam_['minor']['unit']),
                    pa=u.Quantity(beam_['positionangle']['value'], unit=beam_['positionangle']['unit']),
                   )
    elif 'beams' in beam_:
        bdict = beam_['beams']
        if beam_['nStokes'] > 1:
            raise NotImplementedError()
        nbeams = len(bdict)
        assert nbeams == beam_['nChannels']
        stokesidx = '*0'

[...]

PR #738, opened 2021-08-21
Support AIPS-style unit specification

keflavich:
The beam units are specified in the header of the BinTableHDU; we should be reading those when they're legit.  Of course, in the first case we've seen them vary, AIPS uses non-standard unit names.

### DIFF
--- a/spectral_cube/io/fits.py
+++ b/spectral_cube/io/fits.py
@@ -11,6 +11,7 @@ from astropy.wcs import WCS
 from collections import OrderedDict
 from astropy.io.fits.hdu.hdulist import fitsopen as fits_open
 from astropy.io.fits.connect import FITS_SIGNATURE
+from astropy import units as u
 
 import numpy as np
 import datetime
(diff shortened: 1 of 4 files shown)


keflavich:
Solves #737

keflavich:
I don't understand the test failures right now; I think all the generated data should have the correct keyword arguments, but they're not showing up?

keflavich:
@e-koch @astrofrog either of you know why there are failures during teardown?  There are a ton of `ERROR at teardown of` messages, but they don't mean anything to me.  I can't see any code changes that could have caused this, since the tests I edited are copy/pasted from others that do not cause a similar failure.

spectral_cube/io/fits.py
```diff
@@ -96,6 +99,26 @@ def read_data_fits(input, hdu=None, mode='denywrite', **kwargs):
                 if 'BPA' in hdu_item.data.names:
                     beam_table = hdu_item.data
 
+                    # Read the bmaj/bmin units from the header
+                    # (we still assume BPA is degrees because we've never seen an exceptional case)
+                    # this will crash if there is no appropriate header info
```
e-koch, line 104: Should we have an error here if either keyword is missing?

e-koch (approved):
Looks good. We could add an error if keywords are missing from the beam table. However, they _should_ be standard.

Issue #772, opened 2021-11-25
apply_function_parallel_spectral with dask


moellert0:
The function "apply_function_parallel_spectral" works fine for "normal" cubes, but the array describing the spectrum for the applied function always has length 1 when "dask" is used. 

The function which is used with "apply_function_parallel_spectral" gets the spectrum for a certain position as input parameter. Without dask, the length of this input parameter is identical to the number of points along the spectral axis. Using dask, e.g.

with cube.use_dask_scheduler('threads', num_workers=4):  
    cube.apply_function_parallel_spectral(Func)

the array describing the intensity along the spectral axis, which is send to "Func" has always length 1 although the spectral axis of "cube" has e.g. 90 points. Without dask, the array, which is send to "Func" has a length of e.g. 90.

keflavich:
Can you provide a specific example so we can reproduce and test this?

moellert0:
Here is an example. The kwargs argument is not read as well:

import astropy.units as u
import numpy as np
from astropy.utils import data
from spectral_cube import SpectralCube


def func(x, add = None):
    print ("\nSpectral axis within func: ", x)
    print ("\nAdd. keyword (should be 1) = ", add)
    y = x * np.nan
    return y


fn = data.get_pkg_data_filename('tests/data/example_cube.fits', 'spectral_cube')
cube = SpectralCube.read(fn, use_dask = True)
print(cube)

print ("\n\n\nSpectral axis of cube[0, :, :] = ", cube[:, 0, 0])

with cube.use_dask_scheduler('threads', num_workers = 4):
    cube.apply_function_parallel_spectral(func, add = 1)

e-koch:
@moellert0 -- This should now be fixed with #773 merged.

PR #773, opened 2021-11-25
Issue 772: dask apply_along_axis missing kwargs

keflavich:
This addresses #772 by passing kwargs and adds a test


keflavich:
This is good by me - whoever reviews, please merge if you think it's good

e-koch:
@keflavich -- Added testing with and without the `accepts_chunks` parameter. `accepts_chunks=False` was missing from testing leading to this issue.

Otherwise, I also agree this is ready to merge.

PR #775, opened 2021-12-08
Refactor: use data * unit rather than u.Quantity(data, unit)

keflavich:
It appears that `u.Quantity(data, unit)` triggers a full file read, which is disastrous as the whole point of spectral-cube is _not_ to trigger unnecessary IO.

My preference for using that construct came from some very old insight gained when performance testing code for `pyradex`.  If you're adding units a lot of time in a loop, `data * unit` is expensive.  But, we're not expecting to add the units many times - that's not a likely bottleneck.  Instead, IO is a bottleneck. It makes more sense to have:

```python
data_with_unit = cube._data * u.K
```
than
```python
data_with_unit = u.Quantity(cube._data, unit=u.K, copy=False)
```
for our use cases; we often want to stitch units onto dask arrays, which doesn't play well with `u.Quantity`'s constructor.

(caveat: I haven't tested this yet, so it's possible there's some astropy version dependence)

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -46,6 +46,7 @@ from .base_class import (BaseNDClass, SpectralAxisMixinClass,
                          DOPPLER_CONVENTIONS, SpatialCoordMixinClass,
                          MaskableArrayMixinClass, MultiBeamMixinClass,
                          HeaderMixinClass, BeamMixinClass,
+                         lazy_quantity
                         )
 from .utils import (cached, warn_slow, VarianceWarning, BeamWarning,
                     UnsupportedIterationStrategyWarning, WCSMismatchWarning,
@@ -265,14 +266,7 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
                                        "does not match the input unit '{1}'."
                                        .format(unit, data.unit))
             else:
-<<<<<<< HEAD
-                pass
-                # there is no need to turn Data into a quantity here; doing so
-                # triggers a read-into-memory if it is dask
-                # data = u.Quantity(data, unit=unit, copy=False)
-=======
-                data = data * unit
->>>>>>> 3e9660d686e2d8a4d741e9555607c005a20b3414
+                data = lazy_quantity(data, unit)
         elif self._unit is not None:
             unit = self.unit
 
@@ -459,7 +453,7 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         if axis is None:
             # return is scalar
             if unit is not None:
-                return out * unit
+                return lazy_quantity(out, unit)
             else:
                 return out
         elif projection and reduce:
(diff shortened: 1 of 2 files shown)


keflavich:
This PR is pretty wrong because `u.Quantity(x, unit)` could sometimes mean "Attach unit `unit` to unitless `x`" and sometimes mean "Convert quantity `x` to unit `unit`"

keflavich:
The actual bottleneck is here:
https://github.com/astropy/astropy/blob/17e281e/astropy/units/quantity.py#L504-L505
i.e., `np.array(dask_data)`

keflavich:
https://github.com/astropy/astropy/issues/8227#issuecomment-951431146 is the same issue

dhomeier:
> i.e., `np.array(dask_data)`

and then
https://github.com/dask/dask/blob/ecddc8df5869679cef5280c869674cbf0d9f60d9/dask/array/core.py#L1551-L1552

> This PR is pretty wrong because `u.Quantity(x, unit)` could sometimes mean "Attach unit `unit` to unitless `x`" and sometimes mean "Convert quantity `x` to unit `unit`"

Apart from this and `(data * unit)` not having a `.unit` method, does the workaround fail anyway to avoid the read?

keflavich:
I don't know!  I think it does, but I didn't get to checking.

keflavich:
Well, OK, I do know that multiplying a dask array by a unit does _not_ trigger the read, so I'm pretty sure this would work within spectral-cube.  I tested by loading a spectral-cube and multiplying `cube._data` (which was a dask array) by `u.K` - that returns a dask array, no problem.  `u.Quantity(cube._data, u.K)` triggers the read.

dhomeier:
`isintance(data, u.Quantity)` would be easy enough to check for separate treatment for a true Quantity, but what if it is already a `(dask_arr * unit)`...?

keflavich:
yes exactly.  I started going down the wrapper route a few hours ago, but I think we need python's formal wrapping tools for that, and then we're getting dangerously close to trying to implement the thing Martin said was very hard.

keflavich:
....I don't know how I managed to flub a merge here.... but I tried to revert the original changeset and go with a wrapper approach, but it's totally WIP and nothing works yet, it's just ideas I started putting on paper

dhomeier:
> what if it is already a `(dask_arr * unit)`...?

Looks like `isinstance(data._meta, u.Quantity)` would work in that case (and also allow to extract the unit), but that would indeed go some way to writing a full wrapper.

spectral_cube/spectral_cube.py
```diff
@@ -171,18 +171,22 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
...
         # must extract unit from data before stripping it
-        if 'BUNIT' in self._meta:
-            self._unit = cube_utils.convert_bunit(self._meta["BUNIT"])
-        elif hasattr(data, 'unit'):
-            self._unit = data.unit
+        if unit is None:
+            if 'BUNIT' in self._meta:
+                self._unit = cube_utils.convert_bunit(self._meta["BUNIT"])
+            elif hasattr(data, 'unit'):
+                self._unit = data.unit
```
dhomeier, line 185:
```suggestion
            elif hasattr(data, 'unit'):
                self._unit = data.unit
            elif hasattr(data, '_meta'):
                self._unit = getattr(data._meta, 'unit', None)
```
might be worth a shot

keflavich:
superceded by #783

Issue #777, opened 2021-12-09
Add support for HighLevelWCS in reproject


keflavich:
spectral-cube's reprojection only works with FITS WCS objects at the moment

```python
from ndcube.wcs.wrappers import ResampledLowLevelWCS
from spectral_cube import SpectralCube
import reproject
from astropy.wcs.wcsapi import HighLevelWCSWrapper

# read whatever random cube I have laying around
cube = SpectralCube.read('G000.00+00.00_H2CO_2pol.fits')[:25,:20,:21]
ww = ResampledLowLevelWCS(cube.wcs, [2,2,1])
hw = HighLevelWCSWrapper(ww)

cube.reproject(hw)
```
yields
```
Traceback (most recent call last):
  File "<ipython-input-39-7934036dabd6>", line 1, in <module>
    cube.reproject(HighLevelWCSWrapper(ww))
  File "/home/adam/repos/spectral-cube/spectral_cube/utils.py", line 49, in wrapper
    return function(self, *args, **kwargs)
  File "/home/adam/repos/spectral-cube/spectral_cube/spectral_cube.py", line 2610, in reproject
    newwcs = wcs.WCS(header)
  File "/home/adam/repos/astropy/astropy/wcs/wcs.py", line 416, in __init__
    for dict_key in orig_header.keys():
AttributeError: 'HighLevelWCSWrapper' object has no attribute 'keys'
```

We should support APE14 HighLevelWCS objects.

The reason reproject takes a header right now is the need to specify an output grid.  However, HighLevelWCS's are supposed to have `array_shape`s defined.

PR #778, opened 2021-12-09
implement APE14 compatibility in reproject

keflavich:
Solves #777 

This is a fairly straightforward PR but needs tests.


keflavich:
This is not even close to straightforward.  I don't think spectral-cube can or should support any non-serializable WCSes.

PR #779, opened 2021-12-09
Attempt to fix Windows build failures on CI

keflavich:
The problems are all invocation errors:
https://github.com/radio-astro-tools/spectral-cube/runs/4475860538?check_suite_focus=true#step:5:255

but it seems to be pointing at a file not closed error:
https://github.com/radio-astro-tools/spectral-cube/runs/4475860538?check_suite_focus=true#step:5:153
```
..\..\spectral_cube\spectral_axis.py .                                   [  0%]
..\..\spectral_cube\tests\test_analysis_functions.py ................... [  1%]
                                                                         [  1%]
..\..\spectral_cube\tests\test_casafuncs.py ......ssssssssssssss         [  2%]
..\..\spectral_cube\tests\test_cube_utils.py .......                     [  2%]
..\..\spectral_cube\tests\test_dask.py ....s...E                         [  3%]
..\..\spectral_cube\tests\test_io.py .E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E.E [  4%]
```
so... let's see if we can fix it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
           - os: windows-latest
             python-version: 3.9
             name: Python 3.9, all dependencies, and dev versions of key dependencies on Windows
-            toxenv: py39-test-all-dev
+            toxenv: py39-test-all-dev-noopenfiles
           - os: ubuntu-latest
             python-version: 3.9
             name: Documentation
--- a/docs/manipulating.rst
+++ b/docs/manipulating.rst
@@ -93,11 +93,11 @@ mask corresponding to the DS9/CRTF region.  `Regions
 parsing.  CRTF regions may also contain spectral cutout information.
 
 This example shows extraction of a subcube from a ds9 region file ``file.reg``.
-`~regions.read_ds9` parses the ds9 file and converts it to a list of
+`~regions.Regions.read` parses the ds9 file and converts it to a list of
 `~regions.Region` objects::
 
     >>> import regions # doctest: +SKIP
-    >>> region_list = regions.read_ds9('file.reg')  # doctest: +SKIP
+    >>> region_list = regions.Regions.read('file.reg')  # doctest: +SKIP
     >>> sub_cube = cube.subcube_from_regions(region_list)  # doctest: +SKIP
 
 This one shows extraction of a subcube from a CRTF region file ``file.crtf``,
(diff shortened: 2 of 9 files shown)


keflavich:
There was a bug introduced in pvextractor that caused gui stuff to fail

keflavich:
Unfortunately I've uncovered a real error (a regression) caused by some change in region handling.  Something about compound region bboxes has changed, breaking our tests

keflavich:
Some things were really broken, including one of our tests.

Both of these tests were incorrectly only getting 1 pixel along the spectral direction, but investigating in ds9 showed that 2 should've been included:
```
                             (('fk5.reg', (slice(None), 1, slice(None))),
                              ('fk5_twoboxes.reg', (slice(None), 1, slice(None))),
```
image

keflavich:
@astrofrog @e-koch I'm requesting reviews from both of you because I made some significant logical changes in the region handling.  I think we had an uncaught error (actually an error in our tests) previously.  It would be good to have an independent look at that if possible

.github/workflows/main.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           - os: windows-latest
             python-version: 3.9
             name: Python 3.9, all dependencies, and dev versions of key dependencies on Windows
-            toxenv: py39-test-all-dev
+            toxenv: py39-test-all-dev-noopenfiles
```
e-koch, line 52: Is the pvextractor gui still causing the open file issue? If not, run one with and without?

e-koch (changes requested):
The changes look good to me. `get_overlap_slices` from the whole cube size is definitely a key fix, I think.

Two changes to add below, since they're not within the current "diff" region:
* Pin `regions>=0.3` (or higher??) in `all`: https://github.com/radio-astro-tools/spectral-cube/blob/master/setup.cfg#L47.
* Let's add a shape/size check to `test_ds9region_new` to cover checking output slice and data values here: https://github.com/radio-astro-tools/spectral-cube/blob/ccf301f50d036a853661c4a15c11f7b6a2e7084e/spectral_cube/tests/test_subcubes.py#L142
```
        assert sc.size == data[result].size
        assert sc.filled_data[:].squeeze().shape == data[result].shape

```

keflavich:
@e-koch you want to make those suggested changes directly on this PR?

e-koch:
@keflavich Done!

e-koch (approved):

PR #782, opened 2021-12-13
Bugfix: change isinstance checks to base class

keflavich:
Right now, `cube*cube` will trigger `_apply_everywhere` instead of `_cube_on_cube_operation` if `cube` is a DaskSpectralCube.  This should fix that mistake.

### DIFF
--- a/spectral_cube/tests/test_dask.py
+++ b/spectral_cube/tests/test_dask.py
@@ -248,9 +248,9 @@ def test_cube_on_cube(filename, request):
 
     # regression test for #782
     # (the regression applies only to CASA .image files)
-    cube = DaskSpectralCube.read(data_adv)
+    cube = DaskSpectralCube.read(dataname)
     assert isinstance(cube, DaskSpectralCube)
-    cube2 = SpectralCube.read(data_adv, use_dask=False)
+    cube2 = SpectralCube.read(dataname, use_dask=False)
     assert not isinstance(cube, DaskSpectralCube)
 
     with patch.object(cube, '_cube_on_cube_operation') as mock:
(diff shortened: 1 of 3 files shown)


keflavich:
Are you guys OK with adding `mock` as a dependency?  It seems super useful for testing things like this,  so I'm happy to add it, but I'm mostly just using examples c&p'd from SO

spectral_cube/tests/test_dask.py
```diff
@@ -239,6 +241,15 @@ def func(x, add=None):
                                    rslt3.filled_data[:].value)
 
 
+def test_cube_on_cube(data_adv):
+    # regression test for #782
+    cube = DaskSpectralCube.read(data_adv)
```
dhomeier, line 246:
```suggestion
def test_cube_on_cube(data_vda_beams):
    # regression test for #782
    cube = DaskSpectralCube.read(data_vda_beams)
```
Did that test really catch the regression? When I ran it on HEAD, it passed, since `DaskSpectralCube.read(data_adv)` is returning a subclass of `SpectralCube`.
Above is returning a `DaskVaryingResolutionSpectralCube`, which is not, so it failed on HEAD, but passes with this PR.

keflavich: You're right!  This is specific to CASA .image files, so I misidentified the error.

keflavich:
Why are we using `BaseSpectralCube` in the CASA reader?
```python
io_registry.register_reader('casa', BaseSpectralCube, load_casa_image)
io_registry.register_reader('casa_image', BaseSpectralCube, load_casa_image)
io_registry.register_identifier('casa', BaseSpectralCube, is_casa_image)
```

keflavich:
Ah, you're right, and that's probably why.  OK, good catch.  I think this regression now covers everything, and it was a mistake to insist on things being subclasses of `SpectralCube`, since `VaryingResolutionSpectralCube` is not a subclass of `SpectralCube`

keflavich:
oh ffs... openfiles _again_

Issue #774, opened 2021-12-08
Workflow question: Arithmetic, etc, w/o loading all into memory


keflavich:
When doing huge cube arithmetic, which is needed in many cases, what workflow do we suggest to avoid OOM errors like this?
```python
Traceback (most recent call last):
  File "<ipython-input-9-bc7d74799a72>", line 3, in <module>
    beam_correct_cube(fn.replace(".image",""))
  File "/orange/adamginsburg/ALMA_IMF/reduction/reduction/cube_finalizing.py", line 41, in beam_correct_cube
    merged = rescale(convmod, epsdict['epsilon'],
  File "/orange/adamginsburg/ALMA_IMF/reduction/reduction/beam_volume_tools.py", line 153, in rescale
    restor = conv_model.unitless + residual*epsilon[:,None,None]
  File "/blue/adamginsburg/adamginsburg/repos/spectral-cube/spectral_cube/spectral_cube.py", line 2245, in __mul__
    return self._apply_everywhere(operator.mul, value)
  File "/blue/adamginsburg/adamginsburg/repos/spectral-cube/spectral_cube/utils.py", line 49, in wrapper
    return function(self, *args, **kwargs)
  File "/blue/adamginsburg/adamginsburg/repos/spectral-cube/spectral_cube/spectral_cube.py", line 925, in _apply_everywhere
    data = function(u.Quantity(self._get_filled_data(fill=self._fill_value),
  File "/orange/adamginsburg/miniconda3/envs/python39/lib/python3.9/site-packages/astropy/units/quantity.py", line 960, in __mul__
    return super().__mul__(other)
  File "/orange/adamginsburg/miniconda3/envs/python39/lib/python3.9/site-packages/astropy/units/quantity.py", line 486, in __array_ufunc__
    result = super().__array_ufunc__(function, method, *arrays, **kwargs)
MemoryError: Unable to allocate 90.5 GiB for an array with shape (1920, 2378, 2661) and data type float64
```

The example code is https://github.com/ALMA-IMF/reduction/blob/1fecfdceb94ab5a673cc4ce7bce9ee82224033f9/reduction/cube_finalizing.py

In brief, it's doing something like:
```python
cube1 = SpectralCube.read(...)
cube2 = SpectralCube.read(...)
cube3 = cube1.convolve_to(beam) + cube2
cube4 = SpectralCube.read(...)
cube5 = cube3 * cube4
cube4.write(...)
cube5.write(...)
```

[...]

PR #783, opened 2021-12-13
Avoid non-lazy reading

keflavich:
This is another independent attempt to solve #774/775

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -941,6 +941,11 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         data_in = self._get_filled_data(fill=self._fill_value)
         data = function(data_in, *args)
 
+        # strip the unit because data_in does not have a unit
+        # (we calculate the appropriate unit above and pass it on below)
+        if hasattr(data, 'unit'):
+            data = data.value
+
         return self._new_cube_with(data=data, unit=new_unit)
 
     @warn_slow
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -921,7 +921,8 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
                                  "cube.  The error was: {0}".format(ex))
 
         # We don't need to convert to a quantity here because the shape check
-        data = function(self._get_filled_data(fill=self._fill_value), *args)
+        data_in = self._get_filled_data(fill=self._fill_value)
+        data = function(data_in, *args)
 
         return self._new_cube_with(data=data)
 
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2277,21 +2277,25 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
             return self._apply_everywhere(operator.truediv, value)
 
     def __floordiv__(self, value):
-        if isinstance(value, BaseSpectralCube):
-            return self._cube_on_cube_operation(operator.floordiv, value)
-        else:
-            # only cube-on-cube division allowed
-            #
-            # we don't support this:
-            # (Pdb) np.array([5,5,5])*u.K // (2*u.K)
-            # <Quantity [2., 2., 2.]>
-            # astropy doesn't support this:
-            # >>> np.array([5,5,5])*u.K // (2*u.Jy)
-            # astropy.units.core.UnitConversionError: Can only apply 'floor_divide' function to quantities with compatible dimensions
-            # >>> np.array([5,5,5])*u.K // (np.array([2])*u.Jy)
-            # astropy.units.core.UnitConversionError: Can only apply 'floor_divide' function to quantities with compatible dimensions
-            raise NotImplementedError("Floor-division (division with truncation) "
-                                      "is not supported.")
+        raise NotImplementedError("Floor-division (division with truncation) "
+                                  "is not supported.")
+        #if isinstance(value, BaseSpectralCube):
+        #    # (Pdb) operator.floordiv(u.K, u.K)
+        #    # *** TypeError: unsupported operand type(s) for //: 'IrreducibleUnit' and 'IrreducibleUnit'
+        #    return self._cube_on_cube_operation(operator.floordiv, value)
+        #else:
+        #    # only cube-on-cube division allowed
+        #    #
+        #    # we don't support this:
+        #    # (Pdb) np.array([5,5,5])*u.K // (2*u.K)
+        #    # <Quantity [2., 2., 2.]>
+        #    # astropy doesn't support this:
+        #    # >>> np.array([5,5,5])*u.K // (2*u.Jy)
+        #    # astropy.units.core.UnitConversionError: Can only apply 'floor_divide' function to quantities with compatible dimensions
+        #    # >>> np.array([5,5,5])*u.K // (np.array([2])*u.Jy)
+        #    # astropy.units.core.UnitConversionError: Can only apply 'floor_divide' function to quantities with compatible dimensions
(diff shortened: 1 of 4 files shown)


keflavich:
`floordiv` was incorrectly being supported in some cases; it shouldn't work!
```python
In [4]: np.array([5])//(2*u.K)
  File "<ipython-input-4-2cff403b3048>", line 1, in <module>
    np.array([5])//(2*u.K)
  File "/orange/adamginsburg/miniconda3/envs/python39/lib/python3.9/site-packages/astropy/units/quantity.py", line 466, in __array_ufunc__
    converters, unit = converters_and_unit(function, method, *inputs)
  File "/orange/adamginsburg/miniconda3/envs/python39/lib/python3.9/site-packages/astropy/units/quantity_helper/converters.py", line 192, in converters_and_unit
    raise UnitConversionError(
UnitConversionError: Can only apply 'floor_divide' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
```

keflavich:
ah shoot no it _should_ be supported.  I have to fix it.

keflavich:
I think this solves a big part of #774/#775 now.  None of the tests explicitly test for data loading - I don't think I know how to do that right now - but they circumvent some of the more problematic cases of `u.Quantity(self)`

spectral_cube/spectral_cube.py
```diff
@@ -903,30 +901,52 @@ def _get_flat_shape(self, axis):
...
         except Exception as ex:
             raise AssertionError("Function could not be applied to a simple "
                                  "cube.  The error was: {0}".format(ex))
 
-        data = function(u.Quantity(self._get_filled_data(fill=self._fill_value),
-                                   self.unit, copy=False),
-                        *args)
+        # We don't need to convert to a quantity here because the shape check
+        data_in = self._get_filled_data(fill=self._fill_value)
+        data = function(data_in, *args)
```
keflavich, line 942: do we need to explicitly delete `data_in` here?

spectral_cube/tests/test_dask.py
```diff
@@ -239,6 +241,15 @@ def func(x, add=None):
                                    rslt3.filled_data[:].value)
 
 
+def test_cube_on_cube(data_adv):
+    # regression test for #782
+    cube = DaskSpectralCube.read(data_adv)
+
+    with patch.object(cube, '_cube_on_cube_operation') as mock:
+        cube * cube
+    mock.assert_called_once()
```
keflavich, line 250: maybe this is where we need an explicit deletion?
```suggestion
    mock.assert_called_once()
    del cube
```

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -344,6 +342,36 @@ def test_apply_everywhere(self, operation, value, data_advs, use_dask):
...
+        assert c1.unit == value.unit
+
+        # append 'o' to indicate that it has been operated on
+        # value.value: the __add__ function explicitly drops the units
+        c1o = c1._apply_everywhere(operation, value.value, check_units=False)
+        d1o = operation(u.Quantity(d1, u.K), value)
+        assert c1o.unit == c1.unit
+        assert c1o.unit == value.unit
+
+        assert np.all(d1o == c1o.filled_data[:])
```
keflavich, line 360: not needed above, but here?
```suggestion
        assert np.all(d1o == c1o.filled_data[:])
        
        del c1
        del c1o
```

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -344,6 +342,36 @@ def test_apply_everywhere(self, operation, value, data_advs, use_dask):
...
+
+    # The workaround for dask data-loading prevents this test from succeeding
+    @pytest.mark.xfail(raises=u.UnitConversionError, strict=True)
+    @pytest.mark.parametrize(('operation', 'value'),
+                             ((operator.div if hasattr(operator,'div') else operator.floordiv, 0.5*u.K),))
+    def test_apply_everywhere_floordivide(self, operation, value, data_advs, use_dask):
+        c1, d1 = cube_and_raw(data_advs, use_dask=use_dask)
+
+        # append 'o' to indicate that it has been operated on
+        c1o = c1._apply_everywhere(operation, value)
```
keflavich, line 370: if this is causing a teardown error, there's nothing we can do about it - this is an xfail

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -344,6 +342,36 @@ def test_apply_everywhere(self, operation, value, data_advs, use_dask):
...
+        d1o = operation(u.Quantity(d1, u.K), value)
+
+        assert np.all(d1o == c1o.filled_data[:])
```
keflavich, line 373: Delete these lines as redundant; the exception should be raised in the previous line
```suggestion
```

keflavich (commented):
I'm trying to track down where the open file is coming from

keflavich:
I had to totally remove the "expected failure" test for floordiv because it was leaving files open - I don't understand that at all.

Otherwise, this goes a long way to avoiding the early reads of full cubes

spectral_cube/tests/test_spectral_cube.py
```diff
@@ -344,6 +342,40 @@ def test_apply_everywhere(self, operation, value, data_advs, use_dask):
...
+    @pytest.mark.parametrize(('operation', 'value'), ((operator.add, 0.5*u.K),
+                                                      (operator.sub, 0.5*u.K),))
+    def test_apply_everywhere_plusminus(self, operation, value, data_advs, use_dask):
+        c1, d1 = cube_and_raw(data_advs, use_dask=use_dask)
+
+        assert c1.unit == value.unit
+
+        # append 'o' to indicate that it has been operated on
+        # value.value: the __add__ function explicitly drops the units
+        c1o = c1._apply_everywhere(operation, value.value, check_units=False)
```
e-koch, line 355: Add a test case with `check_units=True`?

e-koch (approved):
The code looks good to me, though I don't have a good grasp of the `u.Quantity` issue and I think should wait for @astrofrog's review.

e-koch:
Is #782 needed anymore? It looks like all of those changes are here.

keflavich:
This is branched off of #782; that should be merged first

spectral_cube/tests/test_spectral_cube.py
keflavich: `add` is only supposed to use `check_units=False`; multiply/divide use `check_units=True`

keflavich:
The commit history on this is a bit messed up right now; it's overlapped with #782 , which I'd like to resolve first; I'll have to rebase against that

spectral_cube/tests/test_dask.py
```diff
@@ -239,6 +241,40 @@ def func(x, add=None):
...
+    assert isinstance(cube, (DaskSpectralCube, DaskVaryingResolutionSpectralCube))
+    cube2 = SpectralCube.read(dataname, use_dask=False)
+    if 'image' not in filename:
+        # 'image' would be CASA and must be dask
+        assert not isinstance(cube2, (DaskSpectralCube, DaskVaryingResolutionSpectralCube))
+
+    with patch.object(cube, '_cube_on_cube_operation') as mock:
+        cube * cube
+    mock.assert_called_once()
+    del cube
```
dhomeier, line 264:
```suggestion
```
Well, this must certainly not be deleted before you're done with it!

dhomeier (commented):
I ran into some additional `open_files` failures trying to debug this locally, notably in the `oned_`, `twod_` tests the skipped runs were somehow interfering with closing the file; OTOH some of the other explicit removals were no longer necessary and I also got `test_apply_everywhere_floordivide` to pass. The changes are a bit lengthy; I can open another PR against this branch, but if this passes on the CI as is, maybe it's not critical.

PR #788, opened 2021-12-23
Restore `test_apply_everywhere_floordivide`

dhomeier:
Started this as experimental follow-up to #783 to restore the floordivide test, which seems to have failed due to unrelated open file exceptions. Additional open-file failures reported in https://github.com/radio-astro-tools/spectral-cube/pull/783#pullrequestreview-832819170 are fixed in a5f218f, although I could reproduce these only in local to runs under macOS.
Both 6c9a4dc and a5f218f are passing all tests except the unrelated doctest failure; moving the `pytest.skip()` before any file read operations are started in the 2nd commit seems less error-prone in general, but since this did not appear to cause problems in CI, feel free to pick only the 1st one.

### DIFF
--- a/spectral_cube/tests/test_dask.py
+++ b/spectral_cube/tests/test_dask.py
@@ -270,9 +270,6 @@ def test_cube_on_cube(filename, request):
         cube2 * cube
     mock.assert_called_once()
 
-    del cube
-    del cube2
-
 
 if DISTRIBUTED_INSTALLED:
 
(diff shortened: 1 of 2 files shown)


keflavich (approved):
So a couple trivial-looking rearrangements fix the problem?  I still claim no understanding of why files are left open in some cases but not others.  This looks good to merge, though.

dhomeier:
One complication was probably that `open_files` seems to have everything run in a sort of sandbox, making it almost impossible to track the origins of the failures to the respective tests.
The `twod_numpy` tests, as far as I could see, were starting e.g. with `test_twod_numpy[True-sum-slice-0-data_advs]`, which is skipped, but with the file already opened, and that probably interfered with the open-files check for the next `test_twod_numpy[True-sum-cube-0-data_advs]` in my local installation.

PR #796, opened 2022-01-22
Deprecation fix the 2nd: replace `mktemp` with `TemporaryDirectory`

dhomeier:
Another go at replacing the deprecated function for the `save_to_tmp_dir` option, with an attempt to delete the temporary files along with the cube instance.
See however discussion about using `__del__` in https://github.com/radio-astro-tools/spectral-cube/pull/789#pullrequestreview-847262861.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.6.1.dev (unreleased)
 ----------------------
 - Fix memory issue when calling statistics() on FITS cubes. #752
+- Replace deprecated and insecure `mktemp` with `TemporaryDirectory` for
+  ``save_to_tmp_dir`` and add custom deletor to ensure cleanup. #796
 
 0.6 (2021-09-30)
 ----------------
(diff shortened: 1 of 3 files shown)


dhomeier:
Apparently the Windows test is still the only one that runs with `zarr` installed and not skipping the tests. Don't quite understand why, as it's also listed in the `novis` deps.

keflavich (approved):
LGTM, but I think we should hold off on merging for a minute - I'd like to do some additional tests.

In particular, I plan to by-hand load a cube, do some `save_to_tmpdir` operation, verify that the `zarr` is written to disk, delete the cube, verify that the tmpdir has been deleted.  Similarly, same process, but close python w/o deleting tempdir.

Also we should probably figure out the test issue that @dhomeier noted - we should test this on each OS

dhomeier:
> Similarly, same process, but close python w/o deleting tempdir.

Good point, something I've also only tested locally under macOS. Not sure if this can be implemented in the CI – I think `open_files` is kind of testing this, but may not be completely equivalent. And I have no idea about tmpfile handling on Windows in general.

keflavich:
@dhomeier could you rebase this?  I think we're ready to merge even though I never did end up doing that manual test

keflavich:
we merged #789 in favor of this.

PR #801, opened 2022-03-19
Avoid redundant full-cube-passes when region submasking

keflavich:
Region sub-masking has been using `minimal_subcube`, which shouldn't be necessary b/c we already downselect to the minimal bounding box.

Also, it was creating 3D masks when only 2D is needed - the broadcasting might therefore not have been done lazily.

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -2050,7 +2050,8 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
 
         return self.subcube_from_regions(region_list, allow_empty)
 
-    def subcube_from_regions(self, region_list, allow_empty=False):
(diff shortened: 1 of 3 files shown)


keflavich:
Ready for final review

spectral_cube/spectral_cube.py
```diff
@@ -2062,6 +2063,12 @@ def subcube_from_regions(self, region_list, allow_empty=False):
         allow_empty: bool, optional
             If this is False, an exception will be raised if the region
             contains no overlap with the cube. Default is False.
+        minimize : bool
+            Run `minimal_subcube`?  This is mostly redundant, since the
```
e-koch, line 2067:
```suggestion
            Run :meth:`~SpectralCube.minimal_subcube`.  This is mostly redundant, since the
```

e-koch (approved):
Small docs change. Otherwise looks good!

PR #808, opened 2022-05-09
Inherit from ndcube

keflavich:
WIP: let's just see what happens if we try this.


keflavich:
@cadair let's see if we can't get this working.

keflavich:
Tests are _definitely not passing_!!

Open question: `ndcube` uses `__init__` instead of `__new__`.

- Why?
- Should we?
- How do we refactor - just `s/new/init/g` ?

Issue #810, opened 2022-05-10
Update CASA testing versions


e-koch:
We're encountering a python 3.6 setup issue in #809. Given support is now dropped for py37, and CASA >6.4 uses python 3.8, we can probably start dropping the 3.6 testing and support.

keflavich:
:+1: let's drop py36

PR #813, opened 2022-05-13
Remove py36 from testing grid

keflavich:
Solves #810 .

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,9 +31,9 @@ jobs:
             name: Python 3.9 with all non-visualization dependencies (except CASA)
             toxenv: py39-test-novis
           - os: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


e-koch:
@keflavich  -- Also here: https://github.com/radio-astro-tools/spectral-cube/blob/master/.github/workflows/main.yml#L34

.github/workflows/main.yml
```diff
@@ -30,10 +30,6 @@ jobs:
             python-version: 3.9
             name: Python 3.9 with all non-visualization dependencies (except CASA)
             toxenv: py39-test-novis
-          - os: ubuntu-18.04
-            python-version: 3.6
```
e-koch, line 34: @keflavich  This is our only test case with CASA installed. Try bumping to 3.8 to keep it?

PR #817, opened 2022-05-16
Refactor of 813

keflavich:
This might be a cleaner way to do what #813 accomplished.  Let's see.  If it doesn't work, just close it.


keflavich:
ok this approach is just wrong apparently.

PR #818, opened 2022-05-16
Expand setup.cfg requirements

keflavich:
Running tests requires `mock`, which was left out.

`all` should cover tests + docs


astrofrog:
Mock is now in the standard library so we could also just update the imports?

astrofrog (approved):

Issue #819, opened 2022-06-09
Making a pb mask requires hard-coding units that are dimensionless


abulatek:
Ran into an issue when trying to work with a primary beam file in `spectral-cube`. File was imported as follows: 
```
import glob
from spectral_cube import SpectralCube
pb_fn = glob.glob(f"primary_beam.pb") # any .pb file
pb = SpectralCube.read(pb_fn[0], format = 'casa_image')
```
Running `pb.unit` gives a blank cell, confirming that that cube is unitless. However, trying to mask above a certain primary beam level yields an error.
```
mask = pb > 0.8
```
```
ValueError                                Traceback (most recent call last)
<ipython-input-26-6ec680c9e817> in <module>()
----> 1 mask = pb > 0.8

/blue/adamginsburg/abulatek/spectral-cube/spectral_cube/spectral_cube.py in __gt__(self, value)
   2235             The threshold
   2236         """
-> 2237         value = self._val_to_own_unit(value)
   2238         return LazyComparisonMask(operator.gt, value, data=self._data, wcs=self._wcs)
   2239 

/blue/adamginsburg/abulatek/spectral-cube/spectral_cube/spectral_cube.py in _val_to_own_unit(self, value, operation, tofrom, keepunit)
   2224                              " SpectralCubes or Quantities with "
   2225                              "a unit attribute."
-> 2226                              .format(operation=operation, tofrom=tofrom))
   2227 
   2228     def __gt__(self, value):

ValueError: Can only compare cube objects to SpectralCubes or Quantities with a unit attribute.
```
Since `pb.unit` is dimensionless, this should work. What we have to do instead is:
```
mask = pb > 0.8 * pb.unit
```

keflavich:
The fundamental problem here is that we should let `u.dimensionless_unscaled` go through w/o attempting to convert.

PR #820, opened 2022-06-10
Fix issue 819: allow comparison to unitless scalar numbers

keflavich:
Bugfix for #819.   Unitless cubes (those with `dimensionless_unscaled` units) should be comparable to numbers and arrays directly.


keflavich:
New unassociated error:
https://github.com/radio-astro-tools/spectral-cube/runs/6832968047?check_suite_focus=true#step:5:261
Looks like the windows version lost accuracy?  I'll open a separate issue about this.

e-koch (approved):

PR #823, opened 2022-06-12
WIP fix for 821: test spectral smooth on windows

keflavich:
This is a WIP: no fix has been put in place, I just made the tests more granular so we can see where the failure occurs.

Sidenote: It was _extremely_ difficult to set up the testing environment to enable me to debug this.  I had to copy-paste line-by-line code from `conftest.py` and `test_regrid` to get the cubes put in place and then loaded.  This is an immense headache and makes me hate working with pytest.  Is there any sane way to interactively debug pytest-based functions?

### DIFF
--- a/spectral_cube/tests/test_dask.py
+++ b/spectral_cube/tests/test_dask.py
@@ -12,7 +12,7 @@ from astropy import units as u
 from astropy.utils import data
 
 try:
-    from distributed.utils_test import client, loop, cluster_fixture  # noqa
+    from distributed.utils_test import client, loop, cluster_fixture, cleanup  # noqa
     DISTRIBUTED_INSTALLED = True
 except ImportError:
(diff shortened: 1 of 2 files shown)


keflavich:
This is an upstream error; the kernel is getting populated differently.

keflavich:
OK this isn't an error at all.  Our test needs fixing.

https://github.com/astropy/astropy/pull/13299

keflavich:
I had to rebase this because I branched off the wrong branch, but tests passed; it should be good to go though.

On my local branch, I get an error:
https://github.com/keflavich/spectral-cube/runs/6852738789?check_suite_focus=true#step:5:124
that is caused by an outdated doctestplus.  I don't know why that affects me differently from master.  There may be no solution if this is a real error, though, as it affects python3.6, which is no longer supported by pretty much anyone.

PR #828, opened 2022-07-01
Pass kwargs to reproject

keflavich:
Simple addition, but being done partly to enable access to https://github.com/astropy/reproject/pull/214


e-koch (approved):
LGTM. I don't see a need for an additional test given how simple the change/functionality is.

PR #852, opened 2022-12-01
Bugfix: __len__ test for beams

keflavich:
If you write out a non-varying-spectral-resolution spectrum, it will try to load in as a VRS because the `hasattr(beam, '__len__')` check is true, even though there is no `len`.


e-koch:
Test failures are unrelated to this change. We'll fix in a separate PR (starting in #868)

Issue #881, opened 2023-06-07
converting microns to wave number


alexmaragko:
I have a spectral cube object where the spectral_axis is in microns (u.um) and I would like to update it, or alternatively create a new cube, in wave numbers (1/u.cm).

new_cube = cube.with_spectral_unit(1/u.cm) does not work because it requires an equivalencies=u.spectral(), but with_spectral_unit takes the velocity_convention argument with the ‘relativistic’, ‘radio’, or ‘optical’ options.

cube.spectral_axis.to(1/u.cm, equivalencies=u.spectral()) does the job but I am not sure how to create a new cube that has this as the spectral axis.

keflavich:
The`velocity_convention` keyword should be optional.  `cube.with_spectral_unit` should then just work.  All conversions of velocity coordinates automatically use the `.spectral` equivalency.

If you get an error leaving `velocity_convention` unspecified - that's a legit bug that we should fix, but put in `velocity_convention='radio'` as a placeholder

alexmaragko:
Giving `nu_cube = scube.with_spectral_unit(1/u.cm)` returns `*** KeyError: 'wavenumber'`, and the same error is given even if I include `velocity_convention='radio'`

keflavich:
Ah, then this is a deeper issue; it may be that the FITS standard does not support wavenumber?  I'll have to dig more deeply to be sure

alexmaragko:
Thanks. Is there currently a way to create a new cube using `cube.spectral_axis.to(1/u.cm, equivalencies=u.spectral())` which works as expected?

keflavich:
No, the framework underlying all of what spectral-cube does is the FITS-WCS library.  If FITS doesn't support a unit, we can't serialize the data or compute transformations b/w systems

keflavich:
Ah, but FITS does support wavenumber; this might be a simple fix

PR #882, opened 2023-06-07
add wavn to supported freq types

keflavich:
see #881


keflavich:
Simple addition.  Could use test coverage, but this works and solves an open issue: merging.

PR #885, opened 2023-06-22
fix parameter syntax error introduced in PR 883

tvwenger:
See https://github.com/radio-astro-tools/spectral-cube/issues/884

PR #883, opened 2023-06-21
handle numpy.find_common_type's deprecation

neutrinoceros:
`numpy.find_common_type` is deprecated in numpy 1.25, see https://numpy.org/devdocs/release/1.25.0-notes.html

PR #886, opened 2023-06-22
Fix failing newview

keflavich:
There was some refactor of `Quantity` upstream that we didn't properly accommodate.

This PR incorporates #885 too.


keflavich:
I removed #885 and force-pushed to see if that causes tests to fail (with that included, tests passed!)

keflavich:
We still don't have a test that checks for the deprecation in #883, but that's probably b/c we're not checking for deprecations in tests.

tvwenger (approved):

keflavich:
The CASA test has been hung waiting for a runner for 5 hours now; everything else is green.  I'm merging.

PR #901, opened 2024-02-08
Replace distutils.LooseVersion with packaging.version

olebole:
In Python 3.12, **distutils** does no longer exist. In spectral-cube it is used only for version comparison. This PR replaces it with **packaging.version.parse** resp. **packaging.version.Version**. On places where **LooseVersion** is unused, the import is just removed.

This should make the package compatible to Python 3.12; at least the [tests that we run during the Debian build](https://buildd.debian.org/status/fetch.php?pkg=spectral-cube&arch=all&ver=0.6.5-3&stamp=1707355679&raw=0) pass.

### DIFF
--- a/spectral_cube/np_compat.py
+++ b/spectral_cube/np_compat.py
@@ -1,7 +1,6 @@
 from __future__ import print_function, absolute_import, division
 
 import numpy as np
-from distutils.version import LooseVersion
 
 def allbadtonan(function):
     """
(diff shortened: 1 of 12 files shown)


keflavich (approved):

spectral_cube/np_compat.py
```diff
@@ -14,7 +13,7 @@ def f(data, axis=None, keepdims=None):
             result = function(data, axis=axis)
         else:
             result = function(data, axis=axis, keepdims=keepdims)
-        if LooseVersion(np.__version__) >= LooseVersion('1.9.0') and hasattr(result, '__len__'):
+        if hasattr(result, '__len__'):
```
keflavich, line 16: Just checking on this change - are we sure this is OK?  Maybe we've already declared that spectral-cube requires >1.9 somewhere else at this point.... I don't recall

olebole: numpy 1.9.0 was published 2014, and I assumed that it would not run under any of the more modern Python versions anymore. However, I didn't test.

PR #913, opened 2024-06-16
Progressbar: switch to tqdm, add description

keflavich:
As in the title: there are a lot of progressbars scattered throughout spectral-cube, and it can be tricky to track down where they are.  Adding labels should help.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,6 +22,7 @@ install_requires =
     joblib
     casa-formats-io
     packaging
+    tqdm
 
 [options.extras_require]
 test =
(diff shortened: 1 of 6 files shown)


astrofrog (changes requested):
Looks good to me, but can't we just require tqdm as a dependency as it's so lightweight? This will then simplify the code in utils.py and fix the CI.

keflavich:
I'm OK with that.  @e-koch, you good with just making tqdm a requirement?

e-koch:
agreed. as Tom points out, it's lightweight and simplifies things.

keflavich:
I've added tqdm as a dependency and refactored to use it - but the usage of progressbars is still a little non-tqdm-like.

e-koch:
@keflavich can we close this now that #923 is merged?

keflavich:
Yep.

PR #916, opened 2024-06-20
fix for 915: account for different use cases of median

keflavich:


keflavich:
There's a bug in `warn_slow`: it should inspect the function being called to see whether it accepts a `how` keyword.  Presently, it's warning users to use the `how` keyword even if they're using `DaskSpectralCube`s

keflavich:
the last two commits might've been a red herring; I don't think dask is raising the wrong error message.

e-koch (approved):

e-koch:
Merging. Test failures are unrelated.

PR #924, opened 2024-10-12
BUGFIX: Argmax/argmin raywise needs to return int

keflavich:
It was returning floats


e-koch (approved):

e-koch:
Merging. Test failures are unrelated.

PR #963, opened 2025-06-23
Astropy stokescoords support

preshanth:
Astropy now supports stokescoords and I decided to move our stokes coords code to utilize that. Since some of the operations we had have been upstreamed.  I also updated the stokes.rst to show an example that does this. It is adapted from the tests. This should also improve the code coverage overall.  I figured it is time to have stokes class beyond just a placeholder.

### DIFF
--- a/spectral_cube/stokes_spectral_cube.py
+++ b/spectral_cube/stokes_spectral_cube.py
@@ -1,5 +1,6 @@
 import numpy as np
 
+from astropy.coordinates import StokesCoord, custom_stokes_symbol_mapping, StokesSymbol
 from astropy.io.registry import UnifiedReadWriteMethod
 from .io.core import StokesSpectralCubeRead, StokesSpectralCubeWrite
 from .spectral_cube import SpectralCube, BaseSpectralCube
@@ -7,15 +8,6 @@ from . import wcs_utils
 from .masks import BooleanArrayMask, is_broadcastable_and_smaller
 
 __all__ = ['StokesSpectralCube']
-
-VALID_STOKES = ['I', 'Q', 'U', 'V', 'RR', 'LL', 'RL', 'LR', 'XX', 'XY', 'YX', 'YY', 
-                'RX', 'RY', 'LX', 'LY', 'XR', 'XL', 'YR', 'YL', 'PP', 'PQ', 'QP', 'QQ', 
-                'RCircular', 'LCircular', 'Linear', 'Ptotal', 'Plinear', 'PFtotal', 
-                'PFlinear', 'Pangle']
-STOKES_SKY = ['I','Q','U','V']
-STOKES_FEED_LINEAR = ['XX', 'XY', 'YX', 'YY']
-STOKES_FEED_CIRCULAR = ['RR', 'RL', 'LR', 'LL']
-STOKES_FEED_GENERIC = ['PP', 'PQ', 'QP', 'QQ']
 
 
 class StokesSpectralCube(object):
@@ -26,6 +18,30 @@ class StokesSpectralCube(object):
     component-specific masks.
     """
 
+    # Custom stokes symbol mapping for non-standard symbols
+    _custom_stokes_map = {
+        -9: StokesSymbol('RX', 'Custom RX'),
+        -10: StokesSymbol('RY', 'Custom RY'),
+        -11: StokesSymbol('LX', 'Custom LX'),
+        -12: StokesSymbol('LY', 'Custom LY'),
+        -13: StokesSymbol('XR', 'Custom XR'),
+        -14: StokesSymbol('XL', 'Custom XL'),
+        -15: StokesSymbol('YR', 'Custom YR'),
+        -16: StokesSymbol('YL', 'Custom YL'),
+        -17: StokesSymbol('PP', 'Custom PP'),
+        -18: StokesSymbol('PQ', 'Custom PQ'),
+        -19: StokesSymbol('QP', 'Custom QP'),
+        -20: StokesSymbol('QQ', 'Custom QQ'),
+        -21: StokesSymbol('RCircular', 'Custom RCircular'),
+        -22: StokesSymbol('LCircular', 'Custom LCircular'),
+        -23: StokesSymbol('Linear', 'Custom Linear'),
+        -24: StokesSymbol('Ptotal', 'Custom Ptotal'),
+        -25: StokesSymbol('Plinear', 'Custom Plinear'),
+        -26: StokesSymbol('PFtotal', 'Custom PFtotal'),
+        -27: StokesSymbol('PFlinear', 'Custom PFlinear'),
+        -28: StokesSymbol('Pangle', 'Custom Pangle'),
+    }
+
     def __init__(self, stokes_data, mask=None, meta=None, fill_value=None):
 
         self._stokes_data = stokes_data
@@ -34,8 +50,14 @@ class StokesSpectralCube(object):
 
         reference = tuple(stokes_data.keys())[0]
 
-        for component in stokes_data:
+        # Validate and map Stokes components using StokesCoord, with custom mapping
+        try:
+            with custom_stokes_symbol_mapping(self._custom_stokes_map):
+                stokes_coord = StokesCoord(list(stokes_data.keys()))
+        except Exception as e:
+            raise ValueError(f"Invalid Stokes components: {e}")
 
+        for component in stokes_data:
             if not isinstance(stokes_data[component], BaseSpectralCube):
                 raise TypeError("stokes_data should be a dictionary of "
                                 "SpectralCube objects")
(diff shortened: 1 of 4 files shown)


preshanth:
I would appreciate suggestion on if the transform object can be carried out better using spectral_cube more natively. This is a rather naive implementation and I would like some sage advice here @adamginsburg @e-koch. I am tagging @Kitchi so he can verify my transforms and suggestions.

spectral_cube/stokes_spectral_cube.py
```diff
@@ -110,11 +132,13 @@ def wcs(self):
         return self._wcs
 
     def __dir__(self):
-        return self.components + super(StokesSpectralCube, self).__dir__()
+        # Fix: super().__dir__() returns an iterable, convert to list
+        return self.components + list(super(StokesSpectralCube, self).__dir__())
```
preshanth, line 136: This was a fix that was suggested by copilot. Would like a human review.

spectral_cube/stokes_spectral_cube.py
```diff
@@ -34,8 +50,14 @@ def __init__(self, stokes_data, mask=None, meta=None, fill_value=None):
 
         reference = tuple(stokes_data.keys())[0]
 
-        for component in stokes_data:
+        # Validate and map Stokes components using StokesCoord, with custom mapping
+        try:
+            with custom_stokes_symbol_mapping(self._custom_stokes_map):
+                stokes_coord = StokesCoord(list(stokes_data.keys()))
+        except Exception as e:
+            raise ValueError(f"Invalid Stokes components: {e}")
```
keflavich, line 58: Why are we catching a generic exception here?  Let's try not to do that if we can avoid it; this could otherwise unintentionally catch, e.g., interrutps

spectral_cube/stokes_spectral_cube.py
```diff
@@ -45,31 +67,31 @@ def __init__(self, stokes_data, mask=None, meta=None, fill_value=None):
                 raise ValueError("All spectral cubes in stokes_data "
                                  "should have the same WCS")
 
-            if component not in VALID_STOKES:
-                raise ValueError("Invalid Stokes component: {0} - should be one of I, Q, U, V, RR, LL, RL, LR, XX, XY, YX, YY, \
-                                 RX, RY, LX, LY, XR, XL, YR, YL, PP, PQ, QP, QQ, \
-                                 RCircular, LCircular, Linear, Ptotal, Plinear, PFtotal, PFlinear, Pangle".format(component))
+            # Validation is now handled by StokesCoord above
```
keflavich, line 70: is this needed?

spectral_cube/stokes_spectral_cube.py
```diff
@@ -45,31 +67,31 @@ def __init__(self, stokes_data, mask=None, meta=None, fill_value=None):
...
 
             if stokes_data[component].shape != stokes_data[reference].shape:
                 raise ValueError("All spectral cubes should have the same shape")
 
         self._wcs = stokes_data[reference].wcs
         self._shape = stokes_data[reference].shape
+        self._stokes_coord = stokes_coord  # Store StokesCoord instance
```
keflavich, line 77:
```suggestion
        self._stokes_coord = stokes_coord
```
the comment is redundant

spectral_cube/stokes_spectral_cube.py
keflavich: this seems likely to be correct

spectral_cube/stokes_spectral_cube.py
```diff
@@ -193,70 +225,97 @@ def _new_cube_with(self, stokes_data=None,
...
-        U = (XY + YX) / 2
-        V = 1j(XY - YX) / 2
-
-        Circular to Sky
-        I = (RR + LL) / 2
-        Q = (RL + LR) / 2
-        U = 1j*(RL - LR) / 2
-        V = (RR - LL) / 2
+        Transform the Stokes basis of the cube to the one specified if possible.
+        Operates on the underlying data arrays, not SpectralCube objects.
```
keflavich, line 229: should add a note that this is potentially a very expensive operation

spectral_cube/stokes_spectral_cube.py
```diff
@@ -193,70 +225,97 @@ def _new_cube_with(self, stokes_data=None,
...
+                'RL': make_cube(RL, 'I'),
+                'LR': make_cube(LR, 'I'),
+                'LL': make_cube(LL, 'I'),
+            }
+        elif stokes_basis == "Sky" and self.stokes_type == "SKY_STOKES":
+            data = self._stokes_data
+        elif stokes_basis == "Linear" and self.stokes_type == "FEED_LINEAR":
+            data = self._stokes_data
+        elif stokes_basis == "Circular" and self.stokes_type == "FEED_CIRCULAR":
+            data = self._stokes_data
```
keflavich, line 311: a bit redundant...

spectral_cube/tests/test_stokes_spectral_cube.py
```diff
@@ -45,21 +51,47 @@ def test_direct_init_invalid_shape(self, use_dask):
...
+            -20: StokesSymbol('QQ', 'Custom QQ'),
+            -21: StokesSymbol('RCircular', 'Custom RCircular'),
+            -22: StokesSymbol('LCircular', 'Custom LCircular'),
+            -23: StokesSymbol('Linear', 'Custom Linear'),
+            -24: StokesSymbol('Ptotal', 'Custom Ptotal'),
+            -25: StokesSymbol('Plinear', 'Custom Plinear'),
+            -26: StokesSymbol('PFtotal', 'Custom PFtotal'),
+            -27: StokesSymbol('PFlinear', 'Custom PFlinear'),
+            -28: StokesSymbol('Pangle', 'Custom Pangle'),
+        }
```
keflavich: is this a repeat of _custom_stokes_map?  We should import it in that case

preshanth: I made that change and pushed in already

spectral_cube/stokes_spectral_cube.py
preshanth: Not needed removing.

spectral_cube/stokes_spectral_cube.py
preshanth: Well it is to return the same object when the requested transform is the same as a the basis of the cube. Should just warn instead ?

keflavich: I mean, this could all be one `elif` block.... it's fine as is, I guess, since it's readable.

spectral_cube/stokes_spectral_cube.py
```diff
@@ -34,8 +50,15 @@ def __init__(self, stokes_data, mask=None, meta=None, fill_value=None):
...
         reference = tuple(stokes_data.keys())[0]
 
-        for component in stokes_data:
+        # Validate and map Stokes components using StokesCoord, with custom mapping
+        # Only catch ValueError from StokesCoord, not all exceptions
+        try:
+            with custom_stokes_symbol_mapping(self._custom_stokes_map):
+                stokes_coord = StokesCoord(list(stokes_data.keys()))
+        except ValueError as e:
+            raise ValueError(f"Invalid Stokes components: {e}")
```
keflavich, line 59: Is the new value error more clear than the original?  If so, fine, but I would kind of expect `StokesCoord` to raise its own entirely valid & useful error message here, such that the try/except is redundant.  But maybe this is actually helpful.

keflavich (approved):

spectral_cube/stokes_spectral_cube.py
```diff
@@ -193,70 +222,96 @@ def _new_cube_with(self, stokes_data=None,
...
+        # Transformations
+        if self.stokes_type == "FEED_LINEAR" and stokes_basis == "Sky":
+            XX = get_data('XX')
+            YY = get_data('YY')
+            XY = get_data('XY')
+            YX = get_data('YX')
+            I = (XX + YY) / 2
+            Q = (XX - YY) / 2
+            U = (XY + YX) / 2
+            V = 1j * (XY - YX) / 2
```
Kitchi, line 250: Following H & B 1996 (https://aas.aanda.org/articles/aas/pdf/1996/07/dst6557.pdf) the sign of V needs to be flipped here, defined as 

```V = -1j * (XY - YX)/2```

File: spectral_cube/stokes_spectral_cube.py [hunk 10]
```diff
@@ -193,70 +222,96 @@ def _new_cube_with(self, stokes_data=None,
...
+                'V': make_cube(V, 'RR'),
+            }
+        elif self.stokes_type == "SKY_STOKES" and stokes_basis == "Linear":
+            I = get_data('I')
+            Q = get_data('Q')
+            U = get_data('U')
+            V = get_data('V')
+            XX = (I + Q) / 2
+            YY = (I - Q) / 2
+            XY = (U + 1j * V) / 2
```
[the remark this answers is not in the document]
preshanth: I was just typing that here. Thanks for catching it.

spectral_cube/stokes_spectral_cube.py
preshanth: Done. Check after the current push and mark both as resolved if you agree.

spectral_cube/stokes_spectral_cube.py
```diff
@@ -193,70 +222,96 @@ def _new_cube_with(self, stokes_data=None,
...
+            }
+        elif self.stokes_type == "SKY_STOKES" and stokes_basis == "Circular":
+            I = get_data('I')
+            Q = get_data('Q')
+            U = get_data('U')
+            V = get_data('V')
+            RR = (I + V) / 2
+            LL = (I - V) / 2
+            RL = (Q + 1j * U) / 2  
+            LR = (Q - 1j * U) / 2
```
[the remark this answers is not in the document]
preshanth: This is what the VLA uses and CASA uses.

Kitchi (approved):
The fixed Stokes changes look good - consistent with HBS and self-consistent.

Issue #967, opened 2025-07-02
read(dask=True) is reading the whole cube into memory


keflavich:
```
In [1]: from spectral_cube import SpectralCube
cube = SpectralCube.read('/orange/adamginsburg/AC
In [2]: cube = SpectralCube.read('/orange/adamginsburg/ACES/mosaics/cubes/CS21_CubeMosaic.fits', use_dask=True)
WARNING: File may have been truncated: actual file length (127034882880) is smaller than the expected size (127034884800) [astropy.io.fits.file]
Killed
```

```
$ python -c "import dask; print(dask.__version__)"
2025.5.1
```

daskless is well-behaved:
```
In [1]: from spectral_cube import SpectralCube

In [2]: cube = SpectralCube.read('/orange/adamginsburg/ACES/mosaics/cubes/CS21_CubeMosaic.fits', use_dask=False)
WARNING: File may have been truncated: actual file length (127034882880) is smaller than the expected size (127034884800) [astropy.io.fits.file]

In [3]: 
```

I'm going to dig into this, but this is disastrously bad behavior.  I wonder if it's trying to figure out the "name" by reading all the data in again...

astrofrog:
@Cadair - wasn't there a change in dask recently that led it to load all the data for you? Or maybe I'm misremembering.

@keflavich can you try and downgrade dask to a version from last year?

Cadair:
Not very recently.

keflavich:
I'm guessing this is the problem: https://github.com/dask/dask/pull/11524.  This is a massive breaking change in dask's behavior.

keflavich:
I've tried to hack around the issue with https://github.com/keflavich/spectral-cube/tree/issue967, but it was not successful; I think da.from_array is being called by the unified io system in a way I cannot access.  I'll try to figure out how to monkeypatch it...

astrofrog:
I'm investigating this - it is indeed a disappointing change on the dask side

astrofrog:
See https://github.com/radio-astro-tools/spectral-cube/pull/968

PR #968, opened 2025-07-03
Avoid copying of arrays by dask with dask>=2024.12.0

astrofrog:
This fixes https://github.com/radio-astro-tools/spectral-cube/issues/967

The approach here is that since dask copies Numpy arrays specifically, we just wrap the array inside a thin wrapper so that dask doesn't know what kind of data underlies it.

### DIFF
--- a/spectral_cube/tests/test_dask.py
+++ b/spectral_cube/tests/test_dask.py
@@ -368,4 +368,7 @@ def test_lazy_data_loading(tmp_path):
 
     finally:
 
-        os.remove(tmp_path / 'cube.fits')
+        try:
+            os.remove(tmp_path / 'cube.fits')
+        except Exception:
+            pass


astrofrog:
Well I used to get out of memory error and now I don't. I'm going to try your MWE now.

astrofrog:
30Gb, but trying yours now

astrofrog:
If it still fails for you, can you double check you have the latest dask version?

keflavich:
ok yeah... running again now... it works.  Let me check against my PR too; it may be that that also solved it and it was drawing from the wrong version of spectral-cube.  Not sure how that's possible, but I'm checking.

keflavich:
I think this does work.  Do you know why your fix work and mine didn't?  I thought the regression was caused by `x.copy`, which I skip, but maybe it's something happening deeper in the dask `Array` initialization?

astrofrog:
I'm not 100% sure why your fix didn't work but I think it's because the copying happens in asarray in the dark spectral cube initialiser.

In any case I wonder if the fix here is a bit simpler as it does not rely on manual initialisation of Array?

keflavich:
Yeah, this does seem like .... fewer lines of code, but it requires a deeper understanding of dask to know that/why it works. That's fine, it just means I'm useless as a maintainer.  

My filesystem appears to be half-broken right now, but this appears to work.  I'll approve it, but I think we want a regression test.

astrofrog:
In terms of regression test it might be possible to do a smaller test if we use something like (note tested it, just an example):

https://pypi.org/project/pytest-resource-usage/

keflavich (approved):
let's merge this but add some kind of regression test (in another issue?)

astrofrog:
@keflavich - shall we merge now this includes the test?

PR #985, opened 2025-11-11
Remove chat link from README

keflavich:
Removed Gitter chat link from README to try to clean out the pypi push failure.  We don't use gitter anyway.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,6 @@
 About
 =====
 
-|Join the chat at https://gitter.im/radio-astro-tools/spectral-cube|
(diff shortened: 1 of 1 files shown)


pllim:
I think you also have to remove password and token from your publish workflow. synphot uses Trusted Publishing, you just have to do something like this:

https://github.com/spacetelescope/synphot_refactor/blob/06a8e55b79a2eaef0a40131274fa32d0ad2264eb/.github/workflows/publish-to-pypi.yml#L116-L117

PR #986, opened 2025-11-11
Enable attestations for PyPI publish action

keflavich:
This is recommended by .... the warnings on pypi publish workflows

Warning: The workflow was run with the 'attestations: true' input, but an explicit password was also set, disabling Trusted Publishing. As a result, the attestations input is ignored.
Warning: Trusted Publishers allows publishing packages to PyPI from automated environments like GitHub Actions without needing to use username/password combinations or API tokens to authenticate with PyPI. Read more: https://docs.pypi.org/trusted-publishers
Warning: A new Trusted Publisher for the currently running publishing workflow can be created by accessing the following link(s) while logged-in as an owner of the package(s):

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -37,5 +37,4 @@ jobs:
           path: dist
       - uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
-          password: ${{ secrets.pypi_password }}
+          attestations: true


e-koch:
Oh I see the problem. @keflavich you need to rebase to update the rtd yaml file: https://github.com/radio-astro-tools/spectral-cube/blob/main/.readthedocs.yml

pllim:
Also do you need a branch manager?

branch manager

e-koch:
> Also do you need a branch manager?
> 
> branch manager

Hired!

keflavich:
CI passes and this is a pypi recommendation / requirement.  Infrastructure.  Merging.

PR #991, opened 2026-01-05
Avoid deprecation warning with dev astropy/astropy>=8

rosteen:
https://github.com/astropy/astropy/pull/18986 was merged 3 days ago and is causing failures in our dev versions CI tests due to `spectral-cube` using `COPY_IF_NEEDED`. I think this is a reasonable fix for now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 
 - Fixed compatibility with dask>=2024.12.0 #968
 
+- Avoid deprecation warning with astropy>=8 #991
+
 0.6.5 (2023-12-05)
 ----------------------
 - Fixed issue with fix from #893 not getting included in the 0.6.4 tag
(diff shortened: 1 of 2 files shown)


keflavich (approved):

e-koch:
Test failure is unrelated and previously known so I'll merge this.

astrofrog:
At this point would it be simpler to check the numpy version and define COPY_IF_NEEDED directly here instead of importing from astropy?

rosteen:
> At this point would it be simpler to check the numpy version and define COPY_IF_NEEDED directly here instead of importing from astropy?

Probably, and anyway I was about to write that despite seeming to work locally, this didn't end up fixing the warning in our CI. So I agree if `spectral-cube` maintainers don't want to pin numpy>=2 yet, something like what you suggest should be done.

PR #997, opened 2026-03-14
Improve documentation for `subcube_from_regions`

harry353:
Rewrote the docstring to be clearer and more complete, expanded the summary and added `Returns` and `Raises` sections. Also added inline comments to explain some other non-obvious steps in the method body. Includes also a minor change for the `minimize` parameter which now defaults to False instead of True.


keflavich (changes requested):
Good clarifications, but changing `minimize=True` to `minimize=False` is a big change in behavior that I'm not sure is warranted.

Also, why all the periods at the ends of comments?  It's not our style in general, though I don't particularly care.

Issue #1000, opened 2026-05-08
NAXIS of moment figure is 0


Firestar-Reimu-Pro:
<img width="1048" height="1266" alt="Image" src="https://github.com/user-attachments/assets/2a238974-3860-4672-a3ab-24932e94d5ec" />

for example, in this cube, the NAXIS of the moment figure should be the same with the original cube: `4321, 2161`

keflavich:
@Firestar-Reimu-Pro see #1001  - does that solve it for you?

Firestar-Reimu:
Yes, this is what I want

e-koch:
Closing with #1001 merged.

PR #1001, opened 2026-05-08
Preserve naxis when subsetting WCS

keflavich:
Solves #1000


keflavich:
@e-koch @astrofrog I think it's right to drop these two tests (cov and latest), but not 100% sure

e-koch:
For cov, removing that line made that test case run again, with a coverage report: https://github.com/radio-astro-tools/spectral-cube/actions/runs/25570976458/job/75103342347?pr=1001

Our tox file must have become out-of-date.

e-koch (approved):

Issue #1003, opened 2026-05-25
parallel=True does not fasten the speed of smoothing


Firestar-Reimu-Pro:
```python
fits_path = "./Test/test.fits"
cube = SpectralCube.read(fits_path)
spectral_width_channels = 5
kernel_spectral = Gaussian1DKernel(spectral_width_channels)
smoothed_cube = cube.spectral_smooth(kernel_spectral, parallel=True, use_memmap=True)
```

This code runs for 11 seconds, with or without `parallel=True`

the test cube is cutted from CRAFTS survey: https://hiverse.zero2x.org/
```
Original Cube: SpectralCube with shape=(995, 221, 481) and unit=K:
 n_x:    481  type_x: RA---CAR  unit_x: deg    range:   112.987500 deg:  124.987500 deg
 n_y:    221  type_y: DEC--CAR  unit_y: deg    range:   -12.987500 deg:   -7.487500 deg
 n_s:    995  type_s: VRAD      unit_s: km / s  range:       99.966 km / s:     300.042 km / s
```

I also tested spatial smoothing and I found that `parallel=True` also does not fasten the speed of smoothing

Firestar-Reimu-Pro:
On the other hand, scipy run this smoothing in **less than 1 second**, WHY?

keflavich:
what scipy command are you running?  a 10x speed difference is pretty severe, so my best guess is that these are simply different operations.

Firestar-Reimu-Pro:
```
from scipy.ndimage import gaussian_filter

raw_array = cube.unmasked_data[:,:,:].value
sigma_spectral = float(spectral_width_channels)
scipy_smoothed_data = gaussian_filter(
    raw_array, 
    sigma=[sigma_spectral, 0, 0], 
    mode='constant', 
    cval=0.0
)
smoothed_cube = cube.__class__(scipy_smoothed_data, wcs=cube.wcs, header=cube.header, meta=cube.meta)
```

If I run:
```
smoothed_data = smoothed_cube.unmasked_data[:,:,:].value
print(f"Difference between astropy and scipy smoothed data: {np.nanmax(np.abs(smoothed_data - scipy_smoothed_data)):.5e}")
```
it will give `Difference between astropy and scipy smoothed data: 7.27596e-12`

Firestar-Reimu:
don't you use scipy for backend?

[...]

PR #1004, opened 2026-05-31
Speed up non-dask spectral-cube smoothing

keflavich:
Fix for #1003: the default path for spectral smoothing is extremely slow compared to in-memory smoothing.  That's kind of the point, we make not-in-memory smoothing possible, but we _also_ make it trivial to cut out small cubes, and we want those to smooth quickly.


keflavich:
@astrofrog this is related to the astropy#19835

PR #1006, opened 2026-07-05
Fix pytest parametrize iterators

keflavich:
This is driven entirely by claude code.  It claims the current CI failures are because of the iterators plus on windows the backend.  Let's see if this fixes CI!


keflavich (commented):
As an infrastructure fix that has gone green, I'm going to declare this OK and merge w/o review.

PR #1007, opened 2026-07-05
Raise test_lazy_data_loading threshold to 1s to fix Windows CI flake

keflavich:
## Description

`test_dask.py::test_lazy_data_loading` asserts that lazily reading an 8 GiB cube completes within 0.3s. Windows CI runners have a systematic ~0.31s floor just to open the file — two independent runs both clocked **exactly 0.309s** as their fastest of three attempts ([example](https://github.com/radio-astro-tools/spectral-cube/actions/runs/28734449568/job/85206055685)) — so the test fails there regardless of retries.

This raises the threshold to 1s, which still cleanly separates lazy reads (<0.35s even on the slowest runners) from an eager load of 8 GiB (many seconds), and keeps the best-of-three retry from #1006 for filesystem hiccups.

Follow-up to #1006; unblocks Windows CI on all currently-open PRs.

### DIFF
--- a/spectral_cube/tests/test_dask.py
+++ b/spectral_cube/tests/test_dask.py
@@ -357,18 +357,23 @@ def test_lazy_data_loading(tmp_path):
 
     try:
 
-        # A slow CI filesystem can push a single read over the threshold, so
-        # take the fastest of three attempts before declaring failure.
+        # Eagerly loading the 8 GiB cube would take many seconds, so 1s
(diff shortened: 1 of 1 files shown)


keflavich:
The other option is we could just delete windows.  Not the tests, mind you, the OS.

keflavich:
OK, windows passes again.  Sorry about the spam guys - I'm going to keep trying to get my PRs' CI to go green.

PR #1008, opened 2026-07-08
mosaic_cubes: slice weight cubes by their own channels (fix NaN stripes)

keflavich:
`mosaic_cubes` (channel method) sliced the per-target-velocity **weight** slab using the **data** cube's `two_closest_channels` indices (and spatial `mincube_slices`), then passed it as an HDU to `reproject_and_coadd`. When a weight cube is on a different spectral grid than its data cube (different nchan / reference), those indices select a weight slab that does **not bracket** the target velocity, so reprojecting the weight HDU extrapolates to 0 -> the pixel has zero footprint and is dropped -> spurious NaN stripes that beat with the data-vs-weight grid offset.

Found in ALMA-ACES giant-cube mosaics: fields whose weight cube had a different nchan than their data cube (e.g. 3840 vs 3839) showed regular ~28-channel NaN blocks in single-coverage regions; flat/broadcast weights (same grid as data) were immune.

Fix: slice each weight cube by its **own** `two_closest_channels(wcube, channel)` so the weight slab always brackets the target velocity. Validated end-to-end on a real ACES cube (channel plane 38.66% -> 39.83% finite; dropped pixels recovered).

### DIFF
--- a/spectral_cube/cube_utils.py
+++ b/spectral_cube/cube_utils.py
@@ -956,11 +956,16 @@ def mosaic_cubes(cubes, spectral_block_size=100, combine_header_kwargs={},
         output_footprint = np.zeros(shape_opt)
     mask_opt = np.zeros(shape_opt[1:])
 
+
     # check that the beams are deconvolvable
     if commonbeam is not None:
-        for cube in cubes:
+        # assemble beams
+        beams = [cube.beam if hasattr(cube, 'beam') else cube.beams.common_beam()
+                 for cube in cubes]
+
+        for beam in beams:
             # this will raise an exception if any of the cubes have bad beams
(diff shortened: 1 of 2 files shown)


keflavich:
I don't like this. The weight cube should be on an identical grid to the data, otherwise it's invalid.  Closing as incorrect.

keflavich:
Closing: the ACES-side root cause is a single field (ao) whose feathered cube has an anomalous spectral WCS (CRPIX3=0, one fewer channel) that offsets its grid from its weight cube. We are handling it upstream (in the ACES mosaic loader) by falling back to a flat/broadcast weight when no matching-grid weight cube exists, rather than changing mosaic_cubes. The same-grid assumption in mosaic_cubes is intentional.

PR #1009, opened 2026-07-15
Preserve singleton axes in arg world results

### LINKED ISSUE #982: argmax_world can fail when cube has size-1 dimensions
If a cube gets sliced down to be size-1 along one dimension, it will create an indexing error:
```python
 cube[:,:1,:].argmax_world(axis=0)
WARNING: PossiblySlowWarning: This function (<function BaseSpectralCube.argmax at 0x15f1003a0>) requires loading the entire cube into memory and may therefore be slow. [spectral_cube.utils]
  Cell In[78], line 1
    cube[:,:1,:].argmax_world(axis=0)
  File ~/repos/spectral-cube/spectral_cube/utils.py:74 in wrapper
    return function(self, *args, **kwargs)
  File ~/repos/spectral-cube/spectral_cube/spectral_cube.py:866 in argmax_world
    return self._argmaxmin_world(axis, 'argmax', **kwargs)
  File ~/repos/spectral-cube/spectral_cube/spectral_cube.py:840 in _argmaxmin_world
    out[~collapsed_mask] = np.nan
  File ~/mambaforge/lib/python3.10/site-packages/astropy/units/quantity.py:1331 in __setitem__
    self.view(np.ndarray).__setitem__(i, self._to_own_unit(value))
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

which is caused by this line:
https://github.com/radio-astro-tools/spectral-cube/blob/97db3c5b865e90f5ba0d433a6205c70fa663c7ff/spectral_cube/cube_utils.py#L553

I'm not sure if it's safe to just remove that line - @e-koch you have the latest blame edit; any idea?  I'm inclined to try removing it, add a regression test, and hope there aren't any side effects.

sapunyangkut:
## Summary

- preserve non-collapsed singleton dimensions in `world_take_along_axis`
- exercise the existing `argmax_world` and `argmin_world` oracle with a size-one cube dimension

## Root cause

`np.take_along_axis` leaves the collapsed axis with length one. Calling `out.squeeze()` removed that axis and every unrelated singleton axis, so `_argmaxmin_world` received a one-dimensional result but indexed it with a two-dimensional collapsed mask. Squeezing only the requested reduction axis keeps the result shape aligned with that mask.

## Validation

- regression test on the exact base: 4 expected failures covering NumPy/Dask and argmax/argmin
- focused regression test: 4 passed
- adjacent `arg_world` tests: 8 passed, 965 deselected
- full `spectral_cube` tests: 1,519 passed, 217 skipped, 17 xfailed
- narrow flake8 syntax/name checks and `git diff --check`
- Rafter local secrets check: no secrets detected

## AI disclosure

This fix was prepared with OpenAI Codex-assisted automation. I reviewed the issue, minimized the patch to the shared root cause and the existing test oracle, and reran the validation above before submission.

Fixes #982.

### DIFF
--- a/spectral_cube/cube_utils.py
+++ b/spectral_cube/cube_utils.py
@@ -550,7 +550,7 @@ def world_take_along_axis(cube, position_plane, axis):
 
     out = np.take_along_axis(world_coords[world_newaxis],
                              position_plane[plane_newaxis], axis=axis)
-    out = out.squeeze()
+    out = out.squeeze(axis=axis)
 
     return out
 
--- a/spectral_cube/tests/test_spectral_cube.py
+++ b/spectral_cube/tests/test_spectral_cube.py
@@ -794,16 +794,18 @@ class TestNumpyMethods(BaseTest):
     @pytest.mark.parametrize('method', ('argmax_world', 'argmin_world'))
     def test_arg_world(self, method, data_adv, use_dask):
         c1, d1 = cube_and_raw(data_adv, use_dask=use_dask)
+        c1 = c1[:, :1, :]
 
         # Pixel operation is same name with "_world" removed.
         arg0_pixel = getattr(c1, method.split("_")[0])(axis=0)
 
         arg0_world = np.take_along_axis(c1.spectral_axis[:, np.newaxis, np.newaxis],
-                                        arg0_pixel[np.newaxis, :, :], axis=0).squeeze()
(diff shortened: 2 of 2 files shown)


keflavich:
Did this actually cause any issues?  I'll let another AI agent review this; it's not high priority.

PR #1012, opened 2026-08-17
Allow multiplicative operations between cubes with non-equivalent units

### LINKED ISSUE #1011: Should be able to perform operations on cubes if one is unitless
Currently, if you try and, say, multiple 2 cubes together where one has units of Jy/beam and the other is unitless this fails, since the units are not equivalent. This should be possible (and astropy.units allows for this), so I think it's just a case of the equivalency checking needing to be relaxed a bit. An example of this is if you're trying to combine a flux/weight cube for linear mosaicking. At the moment I'm having to use a really inelegant workaround where I use protected methods to force the units in, which isn't ideal

keflavich:
## Summary
- Fixes #1011: multiplying/dividing two cubes previously raised `UnitsError` whenever the two cubes' units weren't equivalent, e.g. a `Jy/beam` flux cube and a dimensionless weight cube (a common case for linear mosaicking). Multiplicative operations don't require equivalent units — `astropy.units` already supports combining them algebraically — so this restriction was unnecessarily strict.
- `_cube_on_cube_operation` (which backs `__add__`/`__sub__`/`__mul__`/`__truediv__`/`__pow__` for cube-on-cube arithmetic) no longer performs an upfront blanket `is_equivalent` check. Instead:
  - If the two cubes' units *are* equivalent, the second cube is converted into the first's units before combining data, exactly as before (so e.g. `K * K -> K**2`, `K + K -> K` behavior is unchanged).
  - If the units are *not* equivalent, the raw data is combined directly and the resulting unit is derived by actually performing the operation on `astropy.units.Quantity` instances of shape `(1,1,1)` — `astropy.units` itself enforces that additive operations require equivalent units (raising `UnitsError` otherwise) while multiplicative operations combine units algebraically without that constraint.
- Added regression tests `test_mul_div_unitless_cube` and `test_add_sub_unitless_cube_raises` to `TestArithmetic` in `spectral_cube/tests/test_spectral_cube.py`.
- Added a `CHANGES.rst` entry.

## Test plan
- [x] `pytest spectral_cube/tests/test_spectral_cube.py` — 826 passed, 139 skipped, 12 xfailed (no regressions vs. pre-change baseline of 822 passed)
- [x] Manually verified: `Jy` cube `*`/`/` dimensionless cube now succeeds and returns `Jy`; `+`/`-` between them still correctly raises `UnitsError`; same-unit arithmetic (`K*K`, `K+K`, `K/K`) unchanged
- [ ] CI

### DIFF
--- a/spectral_cube/spectral_cube.py
+++ b/spectral_cube/spectral_cube.py
@@ -961,6 +961,15 @@ class BaseSpectralCube(BaseNDClass, MaskableArrayMixinClass,
         Apply an operation between two cubes.  Inherits the metadata of the
         left cube.
 
+        Units do not need to be equivalent for this to succeed: additive
+        operations (add, subtract) still require equivalent units (e.g., you
+        cannot add a K cube to a Jy/beam cube), but multiplicative operations
+        (multiply, divide, power) are allowed between cubes with different,
+        non-equivalent units, since their units simply combine algebraically
+        (e.g., a Jy/beam cube multiplied by a dimensionless weight cube
+        yields a Jy/beam cube).  ``astropy.units`` enforces this distinction
+        for us via the ``test_result`` computation below.
+
         Parameters
         ----------
         function : function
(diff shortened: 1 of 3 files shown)


spectral_cube/spectral_cube.py
```diff
@@ -961,6 +961,15 @@ def _cube_on_cube_operation(self, function, cube, equivalencies=[], **kwargs):
...
         left cube.
 
+        Units do not need to be equivalent for this to succeed: additive
+        operations (add, subtract) still require equivalent units (e.g., you
+        cannot add a K cube to a Jy/beam cube), but multiplicative operations
+        (multiply, divide, power) are allowed between cubes with different,
+        non-equivalent units, since their units simply combine algebraically
+        (e.g., a Jy/beam cube multiplied by a dimensionless weight cube
+        yields a Jy/beam cube).  ``astropy.units`` enforces this distinction
+        for us via the ``test_result`` computation below.
```
keflavich:
```suggestion
        via the ``test_result`` computation below.
```

spectral_cube/spectral_cube.py
```diff
@@ -961,6 +961,15 @@ def _cube_on_cube_operation(self, function, cube, equivalencies=[], **kwargs):
         Apply an operation between two cubes.  Inherits the metadata of the
         left cube.
 
+        Units do not need to be equivalent for this to succeed: additive
```
keflavich:
```suggestion
        Units do not need to be equivalent for some operations: additive
```

spectral_cube/spectral_cube.py
```diff
@@ -961,6 +961,15 @@ def _cube_on_cube_operation(self, function, cube, equivalencies=[], **kwargs):
...
+        operations (add, subtract) still require equivalent units (e.g., you
```
keflavich:
```suggestion
        operations (add, subtract) require equivalent units (e.g., you
```

e-koch (approved):

PR #1013, opened 2026-08-24
Preserve input dtype in to() and convolution-based operations

### LINKED ISSUE #679: Allow dtype change
@low-sky notes that `SpectralCube.convolve_to` returns a `float64` dtype from what was initially `float32`. We should consider:
* [ ] dtype checks before/after some operations
* [ ] exposing a `with_dtype` state change: e.g., `newcube = cube.with_dtype('float32')`

I don't know how well this would work while avoiding a full read-in of the cube.

### LINKED ISSUE #995: cube.to() does not respect dtype
Continuing on the theme of changing dtypes (#148, #679, #990, ..), it appears `cube.to()` also does not respect the dtype (changed from float32 to float64 in my case).

What would be the best way to control the dtype in this case (to avoid inflating the cubes by a factor 2x)?

```
cube.unit, cube[0].dtype     # (Unit("K"), dtype('float32'))

cube_Jybeam = cube.to(u.Jy/u.beam)

cube_Jybeam.unit, cube_Jybeam[0].dtype     # (Unit("Jy / beam"), dtype('float64'))
```

keflavich:
## Summary

Addresses #995 and the follow-up comment [here](https://github.com/radio-astro-tools/spectral-cube/issues/995#issuecomment-5398039620) noting that convolution-based operations also upcast dtype, not just `cube.to()`.

Several operations were silently upcasting `float32` cubes to `float64`, roughly doubling their memory footprint:

- **`SpectralCube.to()` / `VaryingResolutionSpectralCube.to()` / `LowerDimensionalObject.to()`** (unit conversion, e.g. `cube.to(u.Jy/u.beam)`): these multiply the data by a unit-conversion factor from `cube_utils.bunit_converters`, which always returns a `float64` array. Multiplying a `float32` array by a `float64` array upcasts the result to `float64`. Now the converted data is cast back to the original dtype after the multiplication.
- **`_apply_function_parallel_base`** (the shared implementation behind `apply_function_parallel_spatial`/`apply_function_parallel_spectral`, and therefore behind `SpectralCube.convolve_to`, `spatial_smooth`, `spectral_smooth`, etc.): this always allocated its output buffer with `dtype=float` (i.e. `float64`), regardless of the input cube's dtype, so *any* operation routed through it lost the original dtype even when the underlying function (e.g. `astropy.convolution.convolve`) itself preserved it. This is why the issue comment observed that "any operation that involves convolution" hits this — `convolve_to`'s default `astropy.convolution.convolve_fft` additionally always returns `float64` internally, compounding the problem. The output buffer now matches the input dtype.
- **`VaryingResolutionSpectralCube.convolve_to`** (per-channel-beam cubes) has its own separate output-array allocation (it doesn't go through `apply_function_parallel_spatial`) that had the same `float64`-regardless-of-input bug.
- **`DaskSpectralCube.convolve_to`**'s `convfunc` didn't downcast the result back to the input dtype when `convolve_fft` was explicitly requested (its default, plain `astropy.convolution.convolve`, already preserved dtype, so this only affected the FFT-based path).

`DaskVaryingResolutionSpectralCube.convolve_to` was already dtype-safe (it explicitly allocates a same-dtype output buffer per chunk), so it needed no change.

## Test plan
- Added regression tests to `TestArithmetic`/module scope in `spectral_cube/tests/test_spectral_cube.py`:
  - `test_basic_unit_conversion_preserves_dtype` / `test_basic_unit_conversion_beams_preserves_dtype` — cast a cube to `float32`, convert units, assert the result stays `float32`
  - `test_convolve_to_preserves_dtype` (single-beam, plus explicit `convolve_fft` on the dask path) / `test_convolve_to_multibeam_preserves_dtype` (per-channel beams) — cast a cube to `float32`, convolve to a larger beam, assert the result stays `float32`
- Verified all 8 new tests **fail** without the source changes (confirmed by temporarily reverting just the source files and re-running) and **pass** with them, so they properly exercise the fix rather than being vacuous.
- `pytest spectral_cube/tests/test_spectral_cube.py` — 830 passed, 139 skipped, 12 xfailed (up from 822 passed baseline, no regressions)
- Full `pytest spectral_cube/tests/` (excluding `test_casafuncs.py`, which needs `casatools`, and `test_visualization.py::test_to_glue`, which fails in this environment for an unrelated, pre-existing reason — `glue.app.qt` / Qt backend not installed) — 1565 passed, no regressions
- [ ] CI

### DIFF
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1454,7 +1454,12 @@ class DaskSpectralCube(DaskSpectralCubeMixin, SpectralCube):
 
         # See #631: kwargs get passed within self.apply_function_parallel_spatial
         def convfunc(img, **kwargs):
-            return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape) * beam_ratio_factor
+            # some convolution functions (e.g., convolve_fft) upcast to
+            # float64 regardless of the input dtype (see #995); cast back
+            # to avoid silently inflating the cube in memory
+            convolved = convolve(img, kernel, normalize_kernel=True, **kwargs)
+            result = convolved.reshape(img.shape) * beam_ratio_factor
+            return result.astype(img.dtype, copy=False)
 
         if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
             kwargs['allow_huge'] = self.allow_huge_operations
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -27,6 +27,7 @@ from astropy import convolution
 from astropy import wcs
 
 from . import wcs_utils
+from . import cube_utils
 from .spectral_cube import SpectralCube, VaryingResolutionSpectralCube, SIGMA2FWHM, np2wcs
 from .utils import cached, VarianceWarning, SliceWarning, BeamWarning, SmoothingWarning, BeamUnitsError, PossiblySlowWarning, ArrayWrapper
 from .lower_dimensional_structures import Projection
@@ -1454,15 +1455,19 @@ class DaskSpectralCube(DaskSpectralCubeMixin, SpectralCube):
 
         # See #631: kwargs get passed within self.apply_function_parallel_spatial
         def convfunc(img, **kwargs):
-            # some convolution functions (e.g., convolve_fft) upcast to
-            # float64 regardless of the input dtype (see #995); cast back
-            # to avoid silently inflating the cube in memory
             convolved = convolve(img, kernel, normalize_kernel=True, **kwargs)
             result = convolved.reshape(img.shape) * beam_ratio_factor
+            # belt-and-braces: convolve_fft's dtype/precision are steered via
+            # kwargs below, but guarantee the output dtype matches the input
+            # regardless (see #995)
             return result.astype(img.dtype, copy=False)
 
-        if convolve is convolution.convolve_fft and 'allow_huge' not in kwargs:
-            kwargs['allow_huge'] = self.allow_huge_operations
+        if convolve is convolution.convolve_fft:
+            if 'allow_huge' not in kwargs:
+                kwargs['allow_huge'] = self.allow_huge_operations
+            # avoid the implicit float64 promotion (and associated memory
+            # cost) that convolve_fft otherwise performs on float32 data
+            kwargs = cube_utils.convolve_fft_singleprecision_kwargs(self._data.dtype, kwargs)
 
         return self.apply_function_parallel_spatial(convfunc,
                                                     accepts_chunks=True,
@@ -1615,6 +1620,11 @@ class DaskVaryingResolutionSpectralCube(DaskSpectralCubeMixin, VaryingResolution
                 return out
             else:
                 return img
+
+        if convolve is convolution.convolve_fft:
+            # avoid the implicit float64 promotion (and associated memory
+            # cost) that convolve_fft otherwise performs on float32 data
+            kwargs = cube_utils.convolve_fft_singleprecision_kwargs(self._data.dtype, kwargs)
 
         # Rechunk so that there is only one chunk in the image plane
         cube = self._map_blocks_to_cube(convfunc,
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -486,9 +486,18 @@ class Projection(LowerDimensionalObject, SpatialCoordMixinClass,
         convolution_kernel = \
             beam.deconvolve(self.beam).as_kernel(pixscale)
 
+        if convolve is convolution.convolve_fft:
+            # avoid the implicit float64 promotion (and associated memory
+            # cost) that convolve_fft otherwise performs on float32 data
+            kwargs = cube_utils.convolve_fft_singleprecision_kwargs(self.dtype, kwargs)
+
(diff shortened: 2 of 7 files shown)


e-koch:
@keflavich is this one ready to review?

keflavich:
hmm... yes to review, but I want to review it too.  I think claude wrote it all on its own when I said 'do a pass over all the issues'.

spectral_cube/dask_spectral_cube.py
```diff
@@ -1454,7 +1454,12 @@ def convolve_to(self, beam, convolve=convolution.convolve, **kwargs):
 
         # See #631: kwargs get passed within self.apply_function_parallel_spatial
         def convfunc(img, **kwargs):
-            return convolve(img, kernel, normalize_kernel=True, **kwargs).reshape(img.shape) * beam_ratio_factor
+            # some convolution functions (e.g., convolve_fft) upcast to
+            # float64 regardless of the input dtype (see #995); cast back
+            # to avoid silently inflating the cube in memory
+            convolved = convolve(img, kernel, normalize_kernel=True, **kwargs)
+            result = convolved.reshape(img.shape) * beam_ratio_factor
+            return result.astype(img.dtype, copy=False)
```
keflavich: I think we can do better and have `convolve` avoid promoting the dtype.  iirc, that should be possible.  Investigate whether we can do that to avoid more in-memory use during the operation too.

keflavich: Good call — you were right, this was doable. `astropy.convolution.convolve_fft` exposes a `complex_dtype` kwarg, but numpy's FFT (its default `fftn`/`ifftn`) always computes in complex128 internally regardless of that setting, so `complex_dtype=complex64` alone doesn't actually help. `scipy.fft` does respect the input precision, so swapping in `scipy.fft.fftn`/`ifftn` alongside `complex_dtype=np.complex64` keeps the FFT working arrays at half the size and produces a genuinely float32 result directly (no final `.astype()` needed).

Confirmed with `tracemalloc` on a standalone `convolve_fft` call that this roughly halves peak memory for a float32 input, and that the result stays within float32 precision of the float64 default (~1e-7 relative).

Added this as `cube_utils.convolve_fft_singleprecision_kwargs()` and applied it to all four `convolve_to` implementations (they all hit the same astropy promotion behavior), gated on `kind == 'f' and itemsize == 4` rather than `dtype == np.float32`, since FITS-derived cubes are commonly big-endian (`>f4`) and a strict identity check misses those — I caught that the hard way when a test using real FITS-read data showed 0 `scipy.fft` calls.

Pushed in 413e53e6.

keflavich:
yeah I think this is reviewable  now

spectral_cube/lower_dimensional_structures.py
```diff
@@ -179,7 +179,10 @@ def to(self, unit, equivalencies=[], freq=None):
         factor = cube_utils.bunit_converters(self, unit, equivalencies=equivalencies,
                                              freq=freq)
 
-        converted_array = (self.quantity * factor).value
+        # the unit conversion factor is computed in double precision, which
+        # would otherwise silently inflate, e.g., float32 data to float64
+        # (see #995)
+        converted_array = (self.quantity * factor).value.astype(self.dtype, copy=False)
```
e-koch, line 188: we're missing a LDO test for this

spectral_cube/spectral_cube.py
e-koch, line 4286: Am I reading this write that we'd allow the float32 -> float64 whenever factor is float64, and then convert back to float32? Would that still allow for a 2x memory increase?

If so, the `factor[:, None, None]` should always be much smaller in size and it seems more efficient to change its data type first.

e-koch (changes requested):

spectral_cube/lower_dimensional_structures.py
e-koch: @adamginsburg -- I pushed 2 tests to address this

e-koch:
@keflavich I added the same convolve fix for convolution with 2D LDOs.

spectral_cube/spectral_cube.py
keflavich: I think this comment got attached to the wrong lines.

No, I think if any of the inputs are 64, we promote?

I'll try to parse this out better though...

keflavich:
```suggestion
        # the unit conversion factor may be float64 by default, so if needed demote (or promote) it first
        factor = np.array(factor, dtype=self._data.dtype)
```

spectral_cube/spectral_cube.py
```diff
@@ -4272,11 +4287,16 @@ def to(self, unit, equivalencies=()):
...
-                                       unit=unit)
+            data = self._data*factor[:,None,None]
         else:
-            return self._new_cube_with(data=self._data*factor,
-                                       unit=unit)
+            data = self._data*factor
+
+        # the unit conversion factor is computed in double precision, which
+        # would otherwise silently inflate, e.g., float32 cubes to float64
+        # (see #995)
```
keflavich:
```suggestion
    # possibly redundant: coerce data back into its original dtype
```

spectral_cube/spectral_cube.py
e-koch: Yeah this looks right to me now.

e-koch:
Propagated that last update to all other relevant data types (LDOs, VRSC). I'm happy with the changes now.

e-koch (approved):

PR #1015, opened 2026-08-27
Delegate pixels_per_beam to the radio_beam builtin

### LINKED ISSUE #109: Demonstrate how to calculate pixels-per-beam
There are operations that require converting from Jy/beam or similar units to pixels-per-beam, so we should show how to calculate that in the docs, e.g.:

```python
cube = SpectralCube.read('example.fits')
beam = cube.beam
ppbeam = (cube.beam.sr / (cube.wcs.celestial.proj_plane_pixel_area()*u.deg**2)).decompose()
```
but there's probably a nicer way we can offer.

e-koch:
## Summary
- `BeamMixinClass.pixels_per_beam` and `MultiBeamMixinClass.pixels_per_beam` now delegate to `radio_beam`'s `Beam.pixels_per_beam(wcs)` instead of duplicating the beam-area/pixel-area calculation inline.
- Multi-beam cubes call `Beam.pixels_per_beam` once per beam rather than the vectorized `Beams.pixels_per_beam`. With numpy<2 value-based casting, `Beams.sr` stays float32 for a float32 beam table (as written by CASA), so the vectorized result no longer matched the float64 per-beam factors used by `to()` (`test_multibeam_jpix_checks_array` failed on numpy 1.x by ~7e-8 relative). The per-beam path gives identical float64 values on numpy 1.x and 2.x, and keeps the return type a list as before.
- Bumps the minimum `radio-beam` dependency to `>=0.3.10`, the first release that includes `pixels_per_beam` (radio-astro-tools/radio_beam#150, which closed radio-astro-tools/radio_beam#109). `docs/installing.rst` is updated to match.

This is no longer blocked: radio_beam#150 was merged and released in radio-beam 0.3.10.

Possible upstream follow-up: `Beams.sr` returns float32 on numpy<2 but float64 on numpy>=2 for float32 beam tables.

## Test plan
- [x] CI green, including `py310-test-oldestdeps` (numpy 1.24.0, radio-beam 0.3.10) and `py310-test-casa` (numpy 1.26.4, radio-beam 0.3.10)
- [x] `jpix` tests and beam/unit/convolve/jtok tests pass locally on numpy 1.26.4 and 2.2.6 with radio-beam 0.3.10

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@
   non-equivalent units, e.g. a ``Jy/beam`` cube and a dimensionless
   (unitless) cube.  Additive operations still require equivalent units. #1011
 
+- ``pixels_per_beam`` now delegates to the ``radio_beam`` builtin
+  (``Beam.pixels_per_beam`` / ``Beams.pixels_per_beam``), requiring
+  ``radio-beam>=0.3.8``. (radio-astro-tools/radio_beam#109)
+
 0.6.5 (2023-12-05)
 ----------------------
 - Fixed issue with fix from #893 not getting included in the 0.6.4 tag
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,8 +60,8 @@
   ``downsample_axis``, and ``MaskBase._filled``. #973
 
 - ``pixels_per_beam`` now delegates to the ``radio_beam`` builtin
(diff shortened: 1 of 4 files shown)


e-koch:
Hold off. radio-beam needs a release first. **Update: radio-beam now released**

e-koch:
@keflavich this one's ready now

keflavich (approved):

Issue #673, opened 2020-10-14
Flux conservation when smoothing in units of Jy/beam


ashleythomasbarnes:
When spatially smoothing a cube in units of "Jy/beam", does spectral-spectral cube conserve flux? 

e.g. 

`import radio_beam
from spectral_cube import SpectralCube

beam5 = radio_beam.Beam(major=5*au.arcsec, minor=5*au.arcsec, pa=0*au.deg)
inputfile='some cube.fits'
cubejy = SpectralCube.read(inputfile)
cubejy = cubejy.with_spectral_unit('km/s')
cubejy`

....SpectralCube with shape=(20, 512, 512) and unit=Jy / beam... 

`
cubejy5 = cubejy.convolve_to(beam5)
`

spectral-spectral seems to treat `cubejy5` as though it was in units of K - i.e. not accounting for the 1/beam when smoothing. This then returns an erroneously low flux in the cube.

keflavich:
`Jy/beam` is a unit of surface brightness, just like Kelvin.  The problem you're observing is that the Jy/beam unit is now different: the beam size has changed.

`SpectralCube` presently treats this correctly; the resulting cube has the beam you've convolved to.  You can check this by looking at `original_cube.beam` or `original_cube.beams` and comparing to `cubejy5.beam` - they should be different beams.

You can also check by converting your results to Kelvin.  They should be correct - but they won't be the same in general, since you're averaging regions of different surface brightness.

keflavich:
Ash is right here, actually - we are keeping the cube in units of the original beam; if we update the beam (and we should), we should also be scaling up the flux unit.  This behavior is unique to Jy/beam scaling - for Jy/sr and Kelvin, the behavior is correct.  But for Jy/beam, the `beam` is changing and we haven't accounted for that.

keflavich:
@e-koch you want to look into tackling this?  We need to special-case the `per beam` unit handling.

e-koch:
@keflavich  Will do!

e-koch:
This issue goes a bit deeper: we don't have built in conversion handling for other surface brightness units either (e.g. Jy/beam -> Jy/sr). We can pass equivalencies for simple conversions (https://docs.astropy.org/en/stable/api/astropy.units.equivalencies.beam_angular_area.html#astropy.units.equivalencies.beam_angular_area) but checks for changes in the spatial resolution aren't handled. See #683

e-koch:
[...]

PR #1017, opened 2026-09-01
Fix Projection.convolve_to missing Jy/beam scaling for beam size change

### LINKED ISSUE #1016: [BUG] Convolving a slice does not account for beam variation
If you run `convolve_to` on a cube, it has this code:

```
# Scale Jy/beam units by the change in beam size
if self.unit.is_equivalent(u.Jy / u.beam):
    beam_ratio_factor = (beam.sr / self.beam.sr).value
else:
    beam_ratio_factor = 1.
```

This does not exist in the equivalent `convolve_to` for a cube slice, so the output values can be wrong

keflavich:
**This PR was written by Claude Code.**

## Summary

Fixes #1016.

`Projection.convolve_to` (`spectral_cube/lower_dimensional_structures.py`) — used when convolving a single cube slice/plane, e.g. `cube[0].convolve_to(beam)` — did not scale `Jy/beam` data by the ratio of the new beam's solid angle to the old beam's solid angle:

```python
# Scale Jy/beam units by the change in beam size
if self.unit.is_equivalent(u.Jy / u.beam):
    beam_ratio_factor = (beam.sr / self.beam.sr).value
else:
    beam_ratio_factor = 1.
```

`SpectralCube.convolve_to` and `VaryingResolutionSpectralCube.convolve_to` both already apply this scaling. Since `Jy/beam` is a beam-size-dependent unit, convolving to a different beam without this correction produces values that are off by exactly that ratio — as the issue notes, "the output values can be wrong."

The fix adds the same `beam_ratio_factor` scaling to `Projection.convolve_to`, matching the existing `SpectralCube.convolve_to` implementation.

## Test plan
- Added `test_convolve_to_jybeam_onebeam_slice` to `spectral_cube/tests/test_spectral_cube.py`, using the existing `point_source_5_one_beam` fixture (a `Jy/beam` point-source cube). It checks that convolving a single plane (`cube[0].convolve_to(beam)`) gives the same result as convolving the whole cube and taking the same channel (`cube.convolve_to(beam)[0]`), and that the point source's peak stays ~constant in `Jy/beam`.
- Verified the test **fails** without the source fix (off by exactly the beam-ratio factor — confirmed by temporarily reverting just the source change and re-running) and **passes** with it.
- `pytest spectral_cube/tests/test_spectral_cube.py` — 828 passed, 139 skipped, 12 xfailed (826 baseline + 2 new, no regressions)
- [ ] CI

### DIFF
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -483,8 +483,6 @@ class Projection(LowerDimensionalObject, SpatialCoordMixinClass,
         convolution_kernel = \
             beam.deconvolve(self.beam).as_kernel(pixscale)
 
-        # Scale Jy/beam units by the change in beam size, consistent with
-        # SpectralCube.convolve_to (see #1016)
         if self.unit.is_equivalent(u.Jy / u.beam):
             beam_ratio_factor = (beam.sr / self.beam.sr).value
         else:
--- a/spectral_cube/lower_dimensional_structures.py
+++ b/spectral_cube/lower_dimensional_structures.py
@@ -483,9 +483,16 @@ class Projection(LowerDimensionalObject, SpatialCoordMixinClass,
         convolution_kernel = \
             beam.deconvolve(self.beam).as_kernel(pixscale)
 
+        # Scale Jy/beam units by the change in beam size, consistent with
+        # SpectralCube.convolve_to (see #1016)
+        if self.unit.is_equivalent(u.Jy / u.beam):
+            beam_ratio_factor = (beam.sr / self.beam.sr).value
+        else:
+            beam_ratio_factor = 1.
+
         newdata = convolve(self.value, convolution_kernel,
                            normalize_kernel=True,
-                           **kwargs)
+                           **kwargs) * beam_ratio_factor
 
         self = Projection(newdata, unit=self.unit, wcs=self.wcs,
                           meta=self.meta, header=self.header,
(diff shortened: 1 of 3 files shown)


spectral_cube/tests/test_spectral_cube.py
```diff
@@ -2219,6 +2219,32 @@ def test_convolve_to_jybeam_onebeam(point_source_5_one_beam, use_dask):
     assert cube.unit == u.Jy / u.beam
 
 
+def test_convolve_to_jybeam_onebeam_slice(point_source_5_one_beam, use_dask):
+    # regression test for #1016: convolving a single cube slice/plane
+    # (a Projection) in Jy/beam units must scale by the change in beam
+    # area, exactly like SpectralCube.convolve_to already does -- prior to
+    # the fix, Projection.convolve_to silently omitted this scaling.
```
keflavich:
```suggestion
    # area, exactly like SpectralCube.convolve_to does
```

spectral_cube/lower_dimensional_structures.py
```diff
@@ -483,9 +483,16 @@ def convolve_to(self, beam, convolve=convolution.convolve_fft,
         convolution_kernel = \
             beam.deconvolve(self.beam).as_kernel(pixscale)
 
+        # Scale Jy/beam units by the change in beam size, consistent with
+        # SpectralCube.convolve_to (see #1016)
```
keflavich:
```suggestion
```

e-koch:
Also solves #673

e-koch (approved):

PR #1018, opened 2026-09-01
Build the interpolated mask after reversing the output cube

dylanpulver:
`DaskSpectralCube.spectral_interpolate` built `newbmask` from the still-ascending interpolated array and then reversed only the data, so with a decreasing output grid the mask ended up in the opposite spectral order to the cube it described. The non-dask `SpectralCube.spectral_interpolate` and `BaseOneDSpectrum.spectral_interpolate` both write data and mask through the same `outslice`, so they stay aligned — the dask path is the odd one of the three.

Measured on a 10-channel cube whose first three channels are masked out, regridded onto a grid running the other way (the ordinary frequency-cube-onto-a-velocity-grid case):

```
use_dask=False -> [ 9.  8.  7.  6.  5.  4.  3. nan nan nan]
use_dask=True  -> [nan nan nan nan  5.  4. nan nan nan nan]
```

The mask is the mirror image: pre-reversal `~isnan` is `[F F F F T T T T T T]`, applied to data that has since become `[9 8 7 6 5 4 nan nan nan nan]`. Fix is to move the two lines: reverse first, then build the mask.

Why the suite does not see it. `test_spectral_interpolate_with_mask` negates `CDELT3` and passes `sg[::-1]`, which makes the *output* grid ascending, so `reverse_out` is never `True` with a mask present. `test_spectral_interpolate_reversed` is the only `reverse_out=True` case and asserts `result.spectral_axis` only. Separately, both use `result[:, 0, 0].value`, which does not apply the mask — the new test asserts `filled_data` and `mask.include()`.

Ran, same venv, Python 3.13, macOS arm64, `pytest spectral_cube`: 1542 passed / 198 skipped / 17 xfailed before, 1544 / 198 / 17 after, no failures either way. Mutants, each caught by the new test: (A) revert to building the mask first; (B) drop the mask entirely (`mask=None`) — also breaks three existing tests; (C) keep the mask where it was but reverse it unconditionally. (C) is worth noting: it passes every pre-existing test, which is why the new test covers the ascending grid as well.

Not tested: `--remote-data`, casa/glue/yt IO paths, and anything other than macOS arm64 / Python 3.13.

Left alone: the dask path uses `interp1d(fill_value=..., bounds_error=False)` where the non-dask path uses `np.interp(left=, right=)`, so the two disagree by one channel at the edges of the valid range. Separate change.

Implementation, tests and this description were drafted with Claude Opus 5.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,9 +17,8 @@
   ``SpectralCube.convolve_to`` already applies, which could produce
   incorrect values. #1016
 
-- Fixed ``DaskSpectralCube.spectral_interpolate`` building the output mask
-  before reversing the data, so a decreasing output grid returned a mask in
-  the opposite spectral order to the cube it described. #1018
+- Fixed: ``DaskSpectralCube.spectral_interpolate`` was building the output mask
+  in opposite spectral order from the data when the spectrum was reversed #1018
 
 0.6.5 (2023-12-05)
 ----------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,10 @@
   ``SpectralCube.convolve_to`` already applies, which could produce
   incorrect values. #1016
 
+- Fixed ``DaskSpectralCube.spectral_interpolate`` building the output mask
+  before reversing the data, so a decreasing output grid returned a mask in
+  the opposite spectral order to the cube it described. #1018
+
 0.6.5 (2023-12-05)
 ----------------------
 - Fixed issue with fix from #893 not getting included in the 0.6.4 tag
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1364,8 +1364,6 @@ class DaskSpectralCubeMixin:
         if reverse_out:
             newcube = newcube[::-1, :, :]
 
-        # must come after the reversal above, otherwise the mask is in the
-        # opposite spectral order to the data it describes
         newbmask = BooleanArrayMask(~np.isnan(newcube), wcs=newwcs)
 
         newcube = self._new_cube_with(data=newcube, wcs=newwcs, mask=newbmask,
--- a/spectral_cube/dask_spectral_cube.py
+++ b/spectral_cube/dask_spectral_cube.py
@@ -1361,10 +1361,12 @@ class DaskSpectralCubeMixin:
             else -outdiff.value
         newwcs.wcs.set()
 
-        newbmask = BooleanArrayMask(~np.isnan(newcube), wcs=newwcs)
-
         if reverse_out:
             newcube = newcube[::-1, :, :]
+
+        # must come after the reversal above, otherwise the mask is in the
+        # opposite spectral order to the data it describes
+        newbmask = BooleanArrayMask(~np.isnan(newcube), wcs=newwcs)
 
         newcube = self._new_cube_with(data=newcube, wcs=newwcs, mask=newbmask,
                                       meta=self.meta,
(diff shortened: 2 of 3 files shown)


keflavich:
What's the motivation for this PR?  are you working on spectral-cube analysis and encountered this as a real problem?

The claude-speak is all fine, but I care about _how you found it_.  That's more important than what the test covers.  I'm also not satisfied with the solution, but the test seems fine.

keflavich:
I think this is actually a correct fix to a typo (or "thinko"), but @astrofrog you're in the blame.  It's a simple reverse-order fix.

dylanpulver:
Straight answer: no. I am not doing spectral-cube analysis and I did not hit this in real work. It came out of a deliberate bug hunt, so you should weigh it accordingly.

How I found it. I was looking specifically for places where the dask and non-dask implementations of the same operation disagree, on the theory that one of the two is usually the one that got updated. `spectral_interpolate` has three siblings that do the same job: `spectral_cube.py:3306-3320` and `lower_dimensional_structures.py:874-888` both write the data and the mask through the same `outslice`, and the dask version builds the mask at `:1364` and then reverses the data at `:1366-1367`. Two of the three agree, so I checked what the third one does on a descending grid and got a mask that is the mirror image of the cube it describes.

Then I looked at why the tests do not see it, which is the part I would want to know if it were my repo. `test_spectral_interpolate_with_mask` negates `CDELT3` and also passes `sg[::-1]`, so the output grid comes out ascending and `reverse_out` is never True with a mask present. `test_spectral_interpolate_reversed` is the only `reverse_out=True` case and it asserts `spectral_axis` only. Both also read `result[:,0,0].value`, which does not apply the mask, so even a wrong mask would not show up in the assertion.

On the solution being unsatisfying, I would rather you were right than have this merged as is. Moving the mask construction after the reversal is the smallest change that makes the mask describe the array it is attached to, but the two paths still differ in a second way I did not fix: dask uses `interp1d(fill_value=..., bounds_error=False)` where non-dask uses `np.interp(left=, right=)`, so they disagree by a channel at the edges. If you would rather see the two paths share one implementation, that is a bigger change and I am happy to do it that way instead, or to leave it to you and @astrofrog.

The code, tests and this description were drafted with an AI assistant, which is in the PR description too.

spectral_cube/dask_spectral_cube.py
```diff
@@ -1361,11 +1361,13 @@ def interp_wrapper(y, args):
             else -outdiff.value
         newwcs.wcs.set()
 
-        newbmask = BooleanArrayMask(~np.isnan(newcube), wcs=newwcs)
-
         if reverse_out:
             newcube = newcube[::-1, :, :]
 
+        # must come after the reversal above, otherwise the mask is in the
+        # opposite spectral order to the data it describes
```
keflavich, line 1368:
```suggestion
```

CHANGES.rst
```diff
@@ -17,6 +17,10 @@
   ``SpectralCube.convolve_to`` already applies, which could produce
   incorrect values. #1016
 
+- Fixed ``DaskSpectralCube.spectral_interpolate`` building the output mask
+  before reversing the data, so a decreasing output grid returned a mask in
+  the opposite spectral order to the cube it described. #1018
```
keflavich, line 22:
```suggestion
- Fixed: ``DaskSpectralCube.spectral_interpolate`` was building the output mask
  in opposite spectral order from the data when the spectrum was reversed #1018
```

keflavich (approved):
this is ok but I made changes to the text: we don't need a comment about this in the main code (it's noise) and I could barely parse the description in the changelog.

PR #1024, opened 2026-09-17
Run CASA tests on Python 3.12 and 3.13

e-koch:
- Add `py312-test-casa` and `py313-test-casa` CI jobs.
- Limit the `numpy<2` pin to the py310 CASA job. numpy<2 has no py313 wheels, and casatools 6.7.6 (cp312/cp313) already runs with numpy 2 in radio-beam and casa-formats-io CI.


e-koch:
Minimal change while testing against all casa-supported python versions.

